PR #110, opened 2013-01-10
Added a base class for all FMDatabase errors. 

dodikk:
Added an NSError child as a base class for all FMDatabase errors (both existing and upcoming ones). It should be easier for your users to handle them.

Fixed error domain for SQL validation function.

### DIFF
--- a/FMError.h
+++ b/FMError.h
@@ -1,0 +1,13 @@
+//
+//  FMError.h
+//  fmdb
+//
+//  Created by Oleksandr Dodatko on 1/10/13.
+//
+//
+
+#import <Foundation/Foundation.h>
+
+@interface FMError : NSError
+
+@end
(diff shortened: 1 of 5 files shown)


dodikk:
@ChristianKienle I've just noticed that apple actually uses class hierarchies.

* [NSCFError](https://github.com/a1anyip/iOS-SDK-4.3-Framework-Header-Dump/blob/master/Frameworks/Foundation.framework/NSCFError.h)

* [NSURLError](https://github.com/a1anyip/iOS-SDK-4.3-Framework-Header-Dump/blob/master/Frameworks/Foundation.framework/NSURLError.h)

P.S. I confessed this patch would not be merged. Just thought one might find this fact interesting.

ChristianKienle:
@dodikk 2 in a million

PR #180, opened 2013-08-23
Data Sanitization of enumerable arguments in [FMDatabase executeQuery:withArgumentsInArray:]

groue:
Hi,

This commit provides data sanitization of NSArray, NSSet and NSOrderedSet arguments in [FMDatabase executeQuery:withArgumentsInArray:], and solves two problems:
1. FMDB 2.1 requires users to provide as many question marks in the query as there are elements, and this is tedious when the enumerable object has a variable length.
2. FMDB 2.1 surprises users who do not know well the internals of sqlite, and expect enumerables to be transparently expanded, as in famous ORMs such as ActiveRecord.

For example:

``` objc
// Note the single question mark bound to an NSArray object.
ids = @[@1, @2, @3];
[db executeQuery:@"SELECT * FROM table WHERE id IN (?)" withArgumentsInArray:@[ids]]
```

There are a few caveats, though. They are documented in the method implementation (see attached code).

I hope you'll find this patch useful, and that we'll improve it together.

### DIFF
--- a/src/FMDatabase.h
+++ b/src/FMDatabase.h
@@ -13,6 +13,8 @@
 
     #define FMDBRelease(__v) ([__v release]);
 
+    #define FMDBAutoreleasing
+
 	#define FMDBDispatchQueueRelease(__v) (dispatch_release(__v));
 #else
     // -fobjc-arc
@@ -23,6 +25,8 @@
     #define FMDBReturnRetained(__v) (__v)
 
     #define FMDBRelease(__v)
+
+    #define FMDBAutoreleasing __autoreleasing
 
 	#if TARGET_OS_IPHONE
 		// Compiling for iOS
--- a/src/FMDatabase.h
+++ b/src/FMDatabase.h
@@ -13,8 +13,6 @@
 
     #define FMDBRelease(__v) ([__v release]);
 
-    #define FMDBAutoreleasing
-
 	#define FMDBDispatchQueueRelease(__v) (dispatch_release(__v));
 #else
     // -fobjc-arc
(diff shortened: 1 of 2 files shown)


groue:
Now that I have shipped a [Swift toolkit](https://github.com/groue/GRDB.swift) for SQLite (heavily influenced by fmdb), I have a better understanding of @ccgus' reluctance to accept this. FMDB does not focus at all on reusable SQLite statements, and I missed them. But reusable statements are a big drag for transparent support for array bindings.

@ccgus : it was a strong decision to hide reusable statements from fmdb API. And I have never missed them, even though I knew that such a concept existed. So this decision of yours was probably the good one. Do you remember why you took this path?

ccgus:
Do you mean the decision to "hide" reusable statements, or not accepting this PR?  I don't think I'm hiding reusable statements, I just have them off by default because I think the programmer should know what she/he getting into before turning it on.

As for not accepting the PR - there's just a little "code smell" about it that's tripping warnings in my head.

groue:
> Do you mean the decision to "hide" reusable statements, or not accepting this PR?

Definitely the decision to "hide" reusable statements, which I do not question at all.

> I just have them off by default because I think the programmer should know what she/he getting into before turning it on.

If I understand well your sentence, there is a fmdb API to reusable statements that the programmer can turn on. Is it true?

ccgus:
FMDatabase's - (void)setShouldCacheStatements:(BOOL)value;

groue:
Thanks for the info, @ccgus.

PR #212, opened 2013-12-04
Fixing the result dictionary of an FMResultSet to use the first value of a column if a column name conflict occurs.

cowboygneox:
I ran into an issue when querying data using "SELECT \* FROM TABLE JOIN OTHER TABLE ...". Both tables had a couple conflicting column names, and thus the result dictionary was giving me the values from the joined table, and not the table. I believe that a warning should be thrown at a minimum, as well as default to the first value, rather than the second.

Please let me know if you have any questions. Thanks!

### DIFF
--- a/src/FMResultSet.m
+++ b/src/FMResultSet.m
@@ -131,7 +131,10 @@
             
             NSString *columnName = [NSString stringWithUTF8String:sqlite3_column_name([_statement statement], columnIdx)];
             id objectValue = [self objectForColumnIndex:columnIdx];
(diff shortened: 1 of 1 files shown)


DaiyiLL:
Expect: 
sql:  select a.*, b.* from a left join b on a.id = b.id
resultDictionary: 
{
    "a": {
        "id": xxx,
        "a1": xxx,
        "a2": xxx,
        "c": xxx
    },
    "b": {
        "id": xxx,
        "b1": xxx,
        "b2": xxx,
        "c": xxx
    }
}

PR #224, opened 2014-01-17
Add a deadlock check to -[FMDatabaseQueue inDatabase:]

mikeash:
Reentrantly calling `-inDatabase:` causes a deadlock which can be kind of annoying to debug, especially if it happens on a background thread and doesn't immediately lock up the UI. This adds a check for reentrant calls and asserts if they happen, causing a nice clean crash instead. The check is not 100% reliable (e.g. two queues which reentrantly call each other will still deadlock but won't be detected) but should catch the common case.

### DIFF
--- a/src/FMDatabaseQueue.m
+++ b/src/FMDatabaseQueue.m
@@ -16,6 +16,13 @@
  something in dispatch_sync
  
  */
+
+/**
+ * A key used to associate the FMDatabaseQueue object with the dispatch_queue_t it uses.
+ * This in turn is used for deadlock detection by seeing if inDatabase: is called on
+ * the queue's dispatch queue, which should not happen and causes a deadlock.
+ */
+static const void * const kDispatchQueueSpecificKey = &kDispatchQueueSpecificKey;
  
 @implementation FMDatabaseQueue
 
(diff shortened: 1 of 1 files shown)


778477:
hi, guys.

`assert` make `Application abort`  doesn't Good。

if `dispatch_get_specific(kXFDBDispatchQueueSpecificKey)` not NULL， means current dispatch queue is _queue


```
 dispatch_block_t syncBlock = ^() {
        
        FMDatabase *db = [self database];
        block(db);
        
        if ([db hasOpenResultSets]) {
            NSLog(@"Warning: there is at least one open result set around after performing [FMDatabaseQueue inDatabase:]");
            
#if defined(DEBUG) && DEBUG
            NSSet *openSetCopy = FMDBReturnAutoreleased([[db valueForKey:@"_openResultSets"] copy]);
            for (NSValue *rsInWrappedInATastyValueMeal in openSetCopy) {
                FMResultSet *rs = (FMResultSet *)[rsInWrappedInATastyValueMeal pointerValue];
                NSLog(@"query: '%@'", [rs query]);
            }
#endif
        }
    };
    
    if(dispatch_get_specific(kXFDBDispatchQueueSpecificKey)){
        syncBlock();
    } else {
        dispatch_sync(_queue, syncBlock);
    }
```

PR #257, opened 2014-05-02
Change header export settings in FMDB and FMDB-IOS targets

aegolden:
These changes cause the FMDB-IOS library target to export headers so projects that depend on FMDB-IOS can find those headers in a natural place under $(BUILT_PRODUCTS_DIR)/usr/local/include.

The second commit updates the header export paths of both FMDB and FMDB-IOS so that headers can be imported like:

```
#import <FMDB/FMDatabase.h>
```

instead of

```
#import <FMDatabase.h>
```


cutz:
@ccgus would you be interested in accepting a pull request like this if it were rebased on master?

ccgus:
@cutz, sure.

Sorry @aegolden, I completely dropped the ball on this one.

PR #284, opened 2014-07-10
Making custom functions

Simon-Zeng:
Hi, I think we can make a little change to the function for making custom sqlite functions. And some very frequently used custom functions are added here. It will save developer's time.


fmdb.podspec
```diff
@@ -5,7 +5,7 @@ Pod::Spec.new do |s|
   s.homepage = 'https://github.com/ccgus/fmdb'
   s.license = 'MIT'
   s.author = { 'August Mueller' => 'gus@flyingmeat.com' }
-  s.source = { :git => 'https://github.com/ccgus/fmdb.git',
```
stephanheilner: The source location should not be your fork.

PR #322, opened 2014-12-14
Changed NSNumber to NSDecimalNumber when retrieving doubles from db

rafek:
This will allow for better precision = less rounding errors.


kaala:
I had a decimal precision problem, the column type is real.
 this fixed the problematic value. but also created precision problem on another value.

I patched by using this code:
```
NSString *str=[self stringForColumnIndex:columnIdx];
returnValue = [NSDecimalNumber decimalNumberWithString:str];
```

igorkulagin:
I would like to vote up this request.
Also NSNumber is superclass of NSDecimalNumber and existing implementations in 3rd-party apps should work.

PR #331, opened 2015-01-14
Don't force custom tokenizer module name

stephanheilner:
Most of the time when using FTS, you'll only need to use a single tokenizer. This change simplifies using a single tokenizer.

@newyankeecodeshop, could you check this out as well?

### DIFF
--- a/FMDB.podspec
+++ b/FMDB.podspec
@@ -1,11 +1,11 @@
 Pod::Spec.new do |s|
   s.name = 'FMDB'
-  s.version = '2.4.1'
+  s.version = '2.5'
   s.summary = 'A Cocoa / Objective-C wrapper around SQLite.'
   s.homepage = 'https://github.com/ccgus/fmdb'
   s.license = 'MIT'
   s.author = { 'August Mueller' => 'gus@flyingmeat.com' }
-  s.source = { :git => 'https://github.com/ccgus/fmdb.git', :tag => 'v2.4.1' }
+  s.source = { :git => 'https://github.com/ccgus/fmdb.git', :tag => 'v2.5' }
   s.requires_arc = true
 
   s.default_subspec = 'standard'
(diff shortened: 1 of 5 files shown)


newyankeecodeshop:
Good idea, but I would make one change. I think the name of the module should always be "fmdb". I'm not sure why we'd want to allow people to register modules under different names. I agree that we could have a registerTokenizer method that takes no name and just uses the default. Thoughts?

stephanheilner:
I'm creating a database that's shared by an app on iOS, Android and Windows Phone. Android and Windows won't have any concept of what fmdb is. That was why I though it would make sense to allow developers to chose the module name.

newyankeecodeshop:
Ah good point. I forgot there are other platforms out there. :-)

stephanheilner:
@ccgus, let me know if you need me to change anything.

ccgus:
I'm just queuing up FMDB stuff at the moment, no worries :)

boshang:
good idea

ccgus:
This patch breaks the tests.  Can you make sure that's fixed and running properly, and send a pull request again?

stephanheilner:
Tests fixed as well as I added a new test for my changes.

stephanheilner:
@ccgus, do you mean I should submit a fresh pull request, or is adding the fixes to the existing one okay?

ccgus:
The existing one is fine.  Why are you changing FMDB's version number though?  I'm going to change that guy back to 2.4.  The "extras" folder isn't part of FMDB's core, so it doesn't get version bumps.

stephanheilner:
I had changed the version because of CocoaPods. Basically, you don't use the master branch with CocoaPods, you use tagged releases, so in order for anyone with CocoaPods to get access to changes, a new tag needs to be created, and then the .podspec file needs to show the latest tag version. Its a CocoaPod-ism.

ccgus:
I've bumped it to 2.5 now.

PR #333, opened 2015-01-17
Correct reported FMDB version

clemensg:
`[FMDatabase FMDBUserVersion]` still reported `2.3` instead of `2.4` as specified in the podspec.


ccgus:
Danke!

Issue #336, opened 2015-01-23
FMDBAdditions in Swift not found in Xcode.


nateuni:
I am porting some existing code over to swift. I am using FMDB in swift and everything has been good so far, but I do not have access to FMDatabaseAdditions methods in my project in spite of having access to everything else (FMResultSet, FMDatabasePool and so on).

1. I have the bridging-header file with the #import "FMDB" statement and it has the includes for all the other FMDB files.
2. I have looked in my build phases and I can see FMDatabaseAdditions.m in the "Compile Sources" and I have moved it to the top of the list.
3. I have cleaned.. 

Still xCode doesn't recognise it or its methods.

Any ideas?

robertmryan:
Perhaps you can show us sample code that illustrates the problem. I use `FMDatabaseAdditions` methods from Swift without incident.

nateuni:
![screen shot 2015-01-23 at 9 27 40 pm](https://cloud.githubusercontent.com/assets/405559/5873071/b5c42982-a346-11e4-9c44-0374baa084ca.png)
![screen shot 2015-01-23 at 9 28 20 pm](https://cloud.githubusercontent.com/assets/405559/5873079/db22178e-a346-11e4-8b16-73755473eb2b.png)
![screen shot 2015-01-23 at 9 31 45 pm](https://cloud.githubusercontent.com/assets/405559/5873121/4f4a5716-a347-11e4-9abc-fd91f0f8c8b3.png)

My understanding is as long as I have the import FMDB.h in the bridging-header then I should be able to access all of its includes (I can see everything else but FMDatabaseAdditions). This code worked perfectly in objective-c. I am not sure what the issue is. 

Note I have added FMDB by dragging (copying) them in, would that be an issue?

robertmryan:
1. Yes, if you have FMDB.h in your bridging header, it should make FMDB methods (including the additions) available from your code. For example, this works fine:

        let db = FMDatabase(path: path)
        db.open()
        
        if let rs = db.executeQuery("select * from test", withArgumentsInArray: nil) {
            while rs.next() {
                let identifier = rs.intForColumn("id")
                let string = rs.stringForColumn("x")

                println("\(identifier): \(string)")
            }
            rs.close()
        }

2. You should not have foundation in your bridging header, though.

[...]

PR #337, opened 2015-01-23
Create variadic renditions of the FMDatabaseAdditions methods

robertmryan:
This is an extension that adds the variadic methods within the Objective-C FMDatabaseAdditions extension.

This addresses the problem identified in issue https://github.com/ccgus/fmdb/issues/336. 

Gus, I wasn't sure about the file name (it's getting a little unwieldy), but I wanted to conform to established patterns. Perhaps a shorter name would be better.

Also, it's worth noting that these methods return implicitly unwrapped optionals. This is a solid Swift pattern, but deviates a little from the Objective-C convention (for fundamental data types such as `int`, `double`, etc., at least). We could return non-optional values for those types if you wanted, but it feels like that's a step in the wrong direction (i.e. the Swift optional lets you differentiate between a `nil` value from `intForQuery` and a zero value).

### DIFF
--- a/CHANGES_AND_TODO_LIST.txt
+++ b/CHANGES_AND_TODO_LIST.txt
@@ -3,6 +3,9 @@ Zip, nada, zilch.  Got any ideas?
 
 If you would like to contribute some code- awesome!  I just ask that you make it conform to the coding conventions already set in here, and to add a couple of tests for your new code to fmdb.m.  And of course, the code should be of general use to more than just a couple of folks.  Send your patches to gus@flyingmeat.com.
 
+2015.01.23
+    Added Swift renditions of the variadic methods of FMDatabaseAdditions.
+
 2014.09.10
     New classes for exposing SQLite's FTS features.  Thanks to Andrew Goodale for the code.
 
--- a/CHANGES_AND_TODO_LIST.txt
+++ b/CHANGES_AND_TODO_LIST.txt
@@ -50,11 +50,11 @@ If you would like to contribute some code- awesome!  I just ask that you make it
 2013.06.04
     Merged in Robert Ryan's comments in .h header files. These comments hopefully make the .h more readable, but just as importantly, can be parsed by [`appledoc`](http://gentlebytes.com/appledoc/) to create HTML documentation or Xcode docsets. <https://github.com/ccgus/fmdb/pull/150>
 
-        - To build that HTML documentation, once you've installed `appledoc`, you issue the command:
+        - To build that HTML documentation, once you've installed `appledoc`, you issue the command [note Jan 2015: the `no-create-docset` needed due to recent changes to `appledoc`]:
 
-             appledoc --project-name FMDB --project-company ccgus --explicit-crossref --no-merge-categories --output ../Documentation --ignore *.m .
+             appledoc --project-name FMDB --project-company ccgus --explicit-crossref --no-merge-categories --no-create-docset --output ../Documentation --ignore *.m .
 
-        - If you want online help integrated right into Xcode, you can issue the command:
+        - If you want online help integrated right into Xcode (which is no longer needed because Xcode now automatically integrates documentation found in the .h files), you can issue the command:
 
              appledoc --project-name FMDB --project-company ccgus --explicit-crossref --merge-categories --install-docset --output --ignore *.m ../Documentation .
 
--- a/Tests/FMDatabaseAdditionsTests.m
+++ b/Tests/FMDatabaseAdditionsTests.m
@@ -101,4 +101,23 @@
     XCTAssertTrue([[self db] userVersion] == 12);
 }
 
+#if SQLITE_VERSION_NUMBER >= 3007017
+- (void)testApplicationID
+{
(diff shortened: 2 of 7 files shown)


robertmryan:
By the way, I contemplated adding this to `FMDatabaseVariadic.swift`, but given that the additions is a separate extension, it felt appropriate to make this a separate extension, too.

robertmryan:
BTW, @ccgus, even if you don't merge this pull request, you may want to take note of https://github.com/robertmryan/fmdb/commit/858216fb488b211efc1677287c29f07f12f4273f because that addresses a completely unrelated problem, whereby `applicationID` and `applicationIDString` appear to have been moved from `FMDatabase` to `FMDatabaseAdditions`, but the test cases were never updated correspondingly. That particular commit addresses that issue, moving a single test from `FMDatabaseTests` to `FMDatabaseAdditionsTests`.

robertmryan:
By the way, the following tests demonstrate the behavior of the Swift variadic functions:

    class FMDatabaseAdditionsVariadicTests: XCTestCase {
        
        var db: FMDatabase!
        var databasePath: String!
        
        override func setUp() {
            super.setUp()
            
            let tempFolder = NSTemporaryDirectory()
            databasePath = tempFolder.stringByAppendingPathComponent("temp.sqlite")
            let fileManager = NSFileManager.defaultManager()
            
            if fileManager.fileExistsAtPath(databasePath) {
                var error: NSError?
                if (!fileManager.removeItemAtPath(databasePath, error: &error)) {
                    XCTFail("Unable to remove old database: \(error?.localizedDescription)")
                }
            }
            
            db = FMDatabase(path: databasePath)
            XCTAssertNotNil(db, "Unable to instantiate FMDatabase")
            
            var success = db.open()
            XCTAssert(success, "Unable to open database")
            
            success = db.executeUpdate("create table test (x integer, y text)")
            XCTAssert(success, "Unable to create table")
            
            success = db.executeUpdate("insert into test (x, y) values (?, ?)", 1, "foo")
            XCTAssert(success, "Unable to insert foo")
            
            success = db.executeUpdate("insert into test (y) values (?)",  "bar")
            XCTAssert(success, "Unable to insert bar")

            success = db.executeUpdate("insert into test (x) values (?)",  3)
            XCTAssert(success, "Unable to insert bar")
        }
        
        override func tearDown() {
            super.tearDown()
            
            let success = db.close()
            XCTAssert(success, "Unable to open database")
            
            var error: NSError?
            if (!NSFileManager.defaultManager().removeItemAtPath(databasePath, error: &error)) {
                XCTFail("Unable to delete \(databasePath)")
            }
            
            db = nil
            databasePath = nil
        }
        
        func testLongForQueryFoundNumeric() {
            let foo = db.longForQuery("select x from test where y = ?", "foo") as Int
            XCTAssertEqual(foo, 1, "foo should have been 1, but was \(foo)")
        }

        func testLongForQueryFoundNull() {
            let bar = db.longForQuery("select x from test where y = ?", "bar")
            XCTAssertNil(bar, "bar should have been 0, but was \(bar)")
        }
        
        func testLongForQueryNotFound() {
            let baz = db.longForQuery("select x from test where y = ?", "baz")
            XCTAssert(baz == nil, "baz should have been nil, but was \(baz)")
        }
        
        func testLongForQueryFoundString() {
            // Note, we're deliberately returning `y`, which will be `"foo"` and using `longForQuery`.
            // SQLite transparently does the conversion for us.

            let nonNumeric = db.longForQuery("select y from test where y = ?", "foo") as Int
            XCTAssertEqual(nonNumeric, 0, "foo should have been 1, but was \(nonNumeric)")
        }
        
        func testStringForQueryFoundString() {
            let foo = db.stringForQuery("select y from test where x = ?", 1) as String
            XCTAssertEqual(foo, "foo", "should have retrieved \"foo\" but retrieved \(foo)")
        }
        
        func testStringForQueryFoundNull() {
            let noValue = db.stringForQuery("select y from test where x = ?", 3)
            XCTAssertNil(noValue, "should have retrieve nil but retrieved \(noValue)")
        }
        
        func testStringForQueryFoundNothing() {
            let noValue = db.stringForQuery("select y from test where x = ?", 999)
            XCTAssertNil(noValue, "should have retrieve nil but retrieved \(noValue)")
        }

        func testStringForQueryFoundNumeric() {
            // Note, we're deliberate returning `x`, which is numeric value of 1, as using `stringForQuery`.
            // SQLite transparently does the conversion for us.

            let string = db.stringForQuery("select x from test where y = ?", "foo") as String
            XCTAssertEqual(string, "1", "should have retrieve \"1\" but retrieved \(string)")
        }
    }

ccgus:
Some weirdness happened with conflicts and such- but… it's all up there now.  Danke!

robertmryan:
Yeah, I saw that but wasn't sure why it happened. I just forked it yesterday. Oh well. Glad you were able to incorporate it without too much effort. 

I'll do pull request on reference docs, too, given that I changed comments in code with the intent of having that percolate through to that reference documentation, too. Note, appledoc uses Apple's new documentation layout, so it will look quite different.

PR #344, opened 2015-02-23
Adding SQLITE_OPEN_URI flag

marcelofabri:
I've updated my SQLite version (as now I need to use SQLCipher), but I was surprised that `FMDatabaseQueue`'s `init`s and `databaseQueueWithPath:` started returning `nil`, because `sqlite3_open_v2` started returning `SQLITE_CANTOPEN`.

Adding `SQLITE_OPEN_URI` to the flags resolved the issue. I know that I can use `initWithPath:flags:`, but I think most people won't realize that this is needed.

### DIFF
--- a/src/fmdb/FMDatabaseQueue.m
+++ b/src/fmdb/FMDatabaseQueue.m
@@ -84,7 +84,7 @@ static const void * const kDispatchQueueSpecificKey = &kDispatchQueueSpecificKey
 - (instancetype)initWithPath:(NSString*)aPath {
     
     // default flags for sqlite3_open
-    return [self initWithPath:aPath flags:SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE];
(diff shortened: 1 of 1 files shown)


ccgus:
So you're adding something like file:/// to the front of your database path?

marcelofabri:
I'm using `absoluteString` on an `NSURL`, which does return `file:///`.

robertmryan:
Obviously, if you used `path` rather than `absoluteString`, this change would be unnecessary. 

I do not think it is correct to make the path rendition of this method accept a URL string. If we wanted one that took `NSURL`, that's a different matter. But methods that take paths generally (never?) accept URL strings and I don't think this should either.

marcelofabri:
@robertmryan I would agree with you if this never worked. However, it was working before updating SQLite and this would be more a convenience.

But, as you said, `FMDatabase` clearly states that it needs a file system path, so I'll update my code to use `path`.

PR #363, opened 2015-04-12
FMDatabaseAdditionsVariadic Fixes for Swift 1.2

Pacoup:
These changes enable compatibility with Swift 1.2 and Xcode 6.3.

### DIFF
--- a/src/extra/Swift extensions/FMDatabaseAdditionsVariadic.swift
+++ b/src/extra/Swift extensions/FMDatabaseAdditionsVariadic.swift
@@ -20,7 +20,7 @@ extension FMDatabase {
         
         if let rs = executeQuery(sql, values!) {
             if rs.next() {
-                let obj = rs.objectForColumnIndex(0) as NSObject
+                let obj = rs.objectForColumnIndex(0) as! NSObject
                 if !(obj is NSNull) {
                     result = completionHandler(rs)
                 }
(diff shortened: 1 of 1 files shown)


robertmryan:
1. In the first commit, https://github.com/Pacoup/fmdb/commit/f9999d63ee9885b3a583f4275633b5d338973531, you propose eliminating `withArgumentsInArray`.

 I'm not sure I'd recommend doing that, because this unnecessarily entails a dependency to `FMDatabaseVariadic.swift`. In practice, that's probably fine (it's unlikely that someone is going to use this FMDatabaseAdditionsVariadic` and not `FMDatabaseVariadic), but it seems unnecessary (and opens up a possibility of some nuisance for some unfortunate users).

 I personally would not suggest introducing this dependency unnecessarily.

2. In the second commit, I agree that if you keep `as NSObject`, you need to force that downcast with `as!`. But if you do that, it only works with Swift 1.2 (not that there are likely too many people still using Swift 1.1).

 I actually submitted another pull request with a more fundamental change, completely removing these `NSObject` dependencies (and removing a bunch of other unnecessary casts, too). See https://github.com/robertmryan/fmdb/commit/f0479b7a638e8f9efe5ab23d5a62984a0eb3674e for more information.

@ccgus That commit was part of a larger pull request (https://github.com/ccgus/fmdb/pull/353) to enable putting FMDB frameworks. Even if you want to defer the commits making FMDB framework-friendly, I'd understand, but the changes in https://github.com/robertmryan/fmdb/commit/f0479b7a638e8f9efe5ab23d5a62984a0eb3674e (and https://github.com/robertmryan/fmdb/commit/9ec2151cff143c20a82440bd8ac72c51f49d7609 lol) probably should be committed so people can use these variadic extensions with both Swift 1.1 and Swift 1.2.

Pacoup:
Oh, I see what you did. Please ignore this PR. It was simply done with suggested fixes in Xcode. https://github.com/robertmryan/fmdb/commit/f0479b7a638e8f9efe5ab23d5a62984a0eb3674e works great.

PR #365, opened 2015-04-18
Add dynamic framework targets for iOS and OSX

insanehunter:


heiberg:
This is good stuff! Any chance it could be merged into the official repo?

PR #371, opened 2015-05-04
Support for Swift1.2

kalpeshdeo:
It gives a compilation error in Swift1.2

### DIFF
--- a/src/extra/Swift extensions/FMDatabaseVariadic.swift
+++ b/src/extra/Swift extensions/FMDatabaseVariadic.swift
@@ -18,8 +18,8 @@ extension FMDatabase {
     ///
     /// :returns: This returns FMResultSet if successful. Returns nil upon error.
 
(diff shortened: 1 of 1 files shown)


robertmryan:
@kalpeshdeo I've also posted a pull request that addresses 1.2 issues. See https://github.com/robertmryan/fmdb/commit/f0479b7a638e8f9efe5ab23d5a62984a0eb3674e. As you can see, I went a slightly different direction (shift completely to Swift objects rather than shifting completely to Cocoa objects), but the intent was the same, to eliminate the Swift 1.2 problems.

@ccgus What do we need to do to fix this? I know you're busy, but I posted my commit over a month ago. The current code does not compile correctly with the current production version of Swift 1.2 included in Xcode 6.3.1.

ccgus:
I'm looking at your pull request now, sorry for the delay.  I've got some comments, and I'll do that over there.

robertmryan:
We’ve long since moved past Swift 1.2 support (now Swift 4), so I’m closing this pull request.

PR #372, opened 2015-05-05
Added async methods for FMDatabaseQueue

davbeck:
These new methods add the option to return immediately and perform the operations on the queue.

`- (void)inDatabase:(void (^)(FMDatabase *db))block async:(BOOL)async`
`- (void)inTransaction:(void (^)(FMDatabase *db, BOOL *rollback))block async:(BOOL)async`
`- (void)inDeferredTransaction:(void (^)(FMDatabase *db, BOOL *rollback))block async:(BOOL)async`
`- (void)inSavePoint:(void (^)(FMDatabase *db, BOOL *rollback))block async:(BOOL)async completion:(void(^)(NSError *error))completion`

### DIFF
--- a/Tests/FMDatabaseQueueTests.m
+++ b/Tests/FMDatabaseQueueTests.m
@@ -62,6 +62,33 @@
     }];
 }
 
+- (void)testAsync
+{
+    XCTestExpectation *queueExpectation = [self expectationWithDescription:@"Async inDatabase operation"];
+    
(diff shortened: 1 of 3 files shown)


ccgus:
Thanks for doing this- I'll be doing something very similar to this for FMDB 3.0- though it'll probably just be another method call instead of adding a flag (which then would call another method internally, with a flag)

PR #379, opened 2015-06-05
add database open with VSF virtual file system

sechel:
Use FMDB in conjunction with, e.g., javerous/SMCryptoFile.

### DIFF
--- a/Tests/FMDatabaseTests.m
+++ b/Tests/FMDatabaseTests.m
@@ -68,6 +68,25 @@
     
 }
 
+- (void)testOpenWithVFS
+{
+    // create custom vfs
+    sqlite3_vfs vfs = *sqlite3_vfs_find(NULL);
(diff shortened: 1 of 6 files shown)


ccgus:
I like this but there should be a could of changes first:

Rename it to:
openWithFlags:(int)flags vfs:(const char*)vfs

(Doesn't seem very cocoa-y to have "and" in there)

Add a similar interface to FMDatabaseQueue, so it's not left out.

And it doesn't seem very likely, but is there anyway to throw in a test or two for this?

sechel:
Ok done. I added tests for opening a db with a custom virtual file system. FMDatabaseQueue is also included now. I also included a shared Scheme for the Tests target.
Greetings.

ccgus:
Danke!

mythodeia:
:+1:

PR #380, opened 2015-06-08
#if defined(DEBUG) && DEBUG compiling error

godera:
replaced by #ifdef DEBUG


ccgus:
What's the compiler error?

phoney:
Actually, IMO both are wrong. Proper use of DEBUG should be 
#if DEBUG
This avoids the problem of #define DEBUG 0
Having said that, the proposed fix doesn't fix any compiler warnings that I can see.

PR #386, opened 2015-06-26
Check if block parameter is not null (after Facebook Infer analyze)

Sigmanet15:
I use `FMDatabase` in many iOS applications and I've analyzed my projects with Facebook Infer (http://fbinfer.com/). The analyze returned two warnings with `block` parameters that could be null.

This little commit just check if `block` variable is not null before calling it.

### DIFF
--- a/src/fmdb/FMDatabase.m
+++ b/src/fmdb/FMDatabase.m
@@ -1315,8 +1315,9 @@ static NSString *FMDBEscapeSavePointName(NSString *savepointName) {
         return err;
     }
     
-	if (block)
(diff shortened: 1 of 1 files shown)


ccgus:
I tell you what- if you use 4 spaces instead of tabs, and put {}'s around the if statement, I'll merge it :)

Sigmanet15:
Haha, sure ! Commit efdf7c5 has spaces and {}.

ccgus:
Danke!

Sigmanet15:
Great ! Your welcome !

PR #398, opened 2015-08-17
Fix custom tokenizers in FTS3

evands:
If the tokenizer shortens the input (as most tokenizers do), the current implementation fails.  This separately determines the range of bytes being replaced and the size of the replacement.


ccgus:
Do you have any tests for this that we can add?

evands:
I do now ;)  See #399

PR #402, opened 2015-08-31
Swift framework carthage

mishagray:
This PR adds OS X 10.10+ & iOS 8.0+ Framework targets that are Carthage friendly, swift 2.0 friendly, and fix the 'sqlite.h" import issues. 

if you want to test, just add:
github "mishagray/fmdb" "swiftFrameworkCarthage"

to your local Cartfile.  Make sure to add 'import FMDB' to any existing swift code.

I've tested this in XCode7b6.  But it should work as-is for Xcode 6.4 (there is no swift 2.0 specific code here!)


robertmryan:
Given that FMDB supports use in framework via Carthage, I'm going to close this pull request.

PR #412, opened 2015-09-28
Make FTS compatible with Swift framework

hermanbanken:
This allows the FTS module to compile too.


newyankeecodeshop:
I'm not sure this is a good idea. The "fts3_tokenizer.h" file is from the SQLite distribution. It definitely needs "sqlite3.h" to be included first; your change is including after the fts3 include. Maybe the issue is the use of "import" instead of "include" for "fts3_tokenizer.h"? At a minimum, I would change this PR to put the include of "sqlite3.h" before "fts3_tokenizer.h".

PR #415, opened 2015-10-09
Allow reentrant calls to FMDatabaseQueue.inDatabase: & FMDatabaseQueu…

kdubb:
…e.beginTransaction:withBlock:

### DIFF
--- a/FMDB.podspec
+++ b/FMDB.podspec
@@ -12,6 +12,7 @@ Pod::Spec.new do |s|
 
   s.subspec 'common' do |ss|
     ss.source_files = 'src/fmdb/FM*.{h,m}'
+    ss.private_header_files = 'src/fmdb/*Private.h'
     ss.exclude_files = 'src/fmdb.m'
   end
 
--- a/README.markdown
+++ b/README.markdown
@@ -319,6 +319,78 @@ if let rs = database.executeQuery("select x, y, z from test", withArgumentsInArr
 database.close()
 ```
 
+Then you can bring it up on the FMDB mailing list by showing your nice and compact FMDBReportABugFunction, or you can report the bug via the github FMDB bug reporter.
+
+## Using FMDB in a framework
+
+1. Clone FMDB repo.
+
+2. Switch to “framework” branch.
+
+3. In your project create a new target (“File” - “New” - “Target…”) and choose “Framework". Call this new target “FMDB” (and make sure to specify language of Objective-C, regardless of the base language for your main project)
+
+4. Copy the contents of the `src/fmdb` directory into your new framework target. **Do not copy the `FMDB.h` file, though.**
+
+5. Edit the `FMDB.h` file in the framework to include whatever headers you want, e.g. 
+
+ ```//
+//  FMDB.h
+//  FMDB
+//
+//  Created by You on 8/2/15.
+//  Copyright (c) 2015 You. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+
+//! Project version number for FMDB.
+FOUNDATION_EXPORT double FMDBVersionNumber;
+
+//! Project version string for FMDB.
+FOUNDATION_EXPORT const unsigned char FMDBVersionString[];
+
+// In this header, you should import all the public headers of your framework using statements like #import <FMDB/PublicHeader.h>
+
+#import "FMDatabase.h"
+#import "FMResultSet.h"
+#import "FMDatabaseAdditions.h"
+#import "FMDatabaseQueue.h"
+```
+
+6. Whichever headers you include in this `FMDB.h` must be changed to “public” headers as discussed here: http://stackoverflow.com/a/24317456/1271826
+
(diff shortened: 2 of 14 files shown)


ccgus:
I think reentrant calls are a pretty bad idea.  That's why the assert is there.

kdubb:
Any examples of why they are a bad idea? dispatch_sync really seems to be the limiter here. Although I do love the GCD paradigm, reentrant locks, like mutexes, have existed forever.

Without this writing DB code that delegates to other DB code becomes a really big headache.  Our usage is simple we load objects that in turn intelligently decide when to cache their dependent objects; the dependents can be loaded explicitly as well.  The easy solution is reentrance.  The complicated solution involves writing specialized functions that load without using the inDatabase/inTransaction code; effectively doubling our interface size to achieve the same end goal.

Basically I can see no downside to this and am interested how this could go wrong.

kdubb:
BTW there is no assert in the beginTransaction/inTransaction path. It just hangs.

ccgus:
It's super easy to get yourself into a deadlock with sqlite.  I've seen it a bunch in the past, which is why I threw in the assert.

kdubb:
I cannot see how letting the same queue/thread continue using the database it already has control of would produce a deadlock.  As a matter of fact it seems it only prevents the deadlock your assert is trapping.

These changes have no bearing on deadlocking between queues or locks.  Also there is no race condition as the only way the queue == self condition can succeed is if the current dispatch queue explicitly owns the object we are testing; guaranteed by reentrancy.

Although I'm mindful of cautious multi-threaded programming.  This seems like a really limiting feature born out of an over abundance of caution.

FWIW, this code has been used in a shipping app for over a year with not a single issue have arose.  The app I'm referencing uses dispatch and NSOperation queues for just about everything it does.

kdubb:
Also, I'd like to point out that the assert is not trapping anything to do with sqlite.  It only has to do with dispatch_sync.

ccgus:
The assert keeps you from performing updates while doing updates.  It happened quite frequently in the past, causing deadlocks.

kdubb:
@ccgus It seems we're disconnected on what this change actually does.  Reentrancy only happens on the same thread. So this code is only allowing a single thread to repeatedly access the database.  Essentially this is no different than calling...
```objective-c
[db executeUpdate:@"INSERT one ..."];
[db executeUpdate:@"INSERT two ..."];
```
Don't confuse reentrant with multithreaded.

ccgus:
I know what reentrancy is :)
In the past, people would get deadlocks by somehow managing to do a select inside of a transaction (in different frames).  Some object would start a transaction, and inside that do some sort of update or select, which would then populate some fields of another object, which would then want to update or select from the database before the transaction is finished.  It happened enough that I added the assert to help folks figure out what they are doing wrong.

I still believe that enabling reentrancy is a bad idea- it just has a super bad code smell to it.  If you really want to call updates on the same database object down the stack- then just pass FMDatabase along.

kdubb:
Would you consider this if it was a compile time or runtime selectable feature? Certainly I would leave the default as disabled.

ccgus:
No, sorry.

kdubb:
Your suggestion of "just pass FMDatabase along" does nothing to stop any of the problems you have brought up. By passing the object along you can make all the same mistakes.  Yet citing those as the reasons you refuse a fairly harmless and simple change.

robertmryan:
For whatever it's worth, I have to say that I agree with Gus, that this doesn't really belong in the main repo. This would probably end up being misused by inexperienced developers who don't bother to know the state of the database.

 When I encountered this issue in my own project, passing the `FMDatabase` pointer to the subroutine addressed this very cleanly (making it explicit in my code where I'm "entering" an `inDatabase` or `inTransaction` block and where I have some subsidiary routine that I call when I know I'm already in one if those blocks).

kdubb:
@robertmryan Again, although you make good points about clarity, none of the issues addressed by Gus are actually solved by disallowing reentrancy. Just because you are passing the pointer along doesn't mean you haven't done so with the database in the wrong state.

kdubb:
@ccgus Since this change was unwelcome I'd like you to look over total changes we made to FMDatabasePool and see if they would be welcome. They are in the main branch on 
https://github.com/reTXT/fmdb.

Effectively it turns FMDatabasePool into a WAL enabled Reader/Writer pool.  It uses dispatch queues to manage the synchronization.  We originally had used a dispatch_barrier_sync and a common queue shared between readers and writers (this is how reader/writer locks are effected in GCD).  Eventually we pared it down to just being a pool for readers and serialized writers. Despite all our attempts to produce a SQLITE_BUSY using this code we cannot; at least on iOS 8 and 9.

It does have reentrancy for writers (as readers will just open a new connection in most cases).  If these changes are welcome I might be persuaded to remove it but I think it's especially appropriate in this code since requesting reader and writer access is spelled out so specifically.

Please let me know if we should plan on keeping this segregated in our own code or if you're interested in pulling them in. Maybe renaming it is appropriate as well (e.g. FMDatabaseWALPool)

PR #416, opened 2015-10-10
Allow selecting NSDate epoch

kdubb:
You can select between the 1970 and reference epoch at runtime .  Works much like the date formatter setting including adding FMDatabase.doubleFromDate & FMDatabase.dateFromDouble methods.

### DIFF
--- a/src/fmdb/FMDatabase.h
+++ b/src/fmdb/FMDatabase.h
@@ -90,6 +90,8 @@ typedef int(^FMDBExecuteStatementsCallbackBlock)(NSDictionary *resultsDictionary
     NSMutableSet        *_openFunctions;
 
     NSDateFormatter     *_dateFormat;
+    SEL                 _dateInitializer;
+    SEL                 _dateExtractor;
 }
 
 ///-----------------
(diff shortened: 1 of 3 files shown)


robertmryan:
I might suggest method names that don't start with `set`, as that's a convention for setters methods. Perhaps `useDateEpoch1970` and `useDateEpochReferenceDate` or something like that.

robertmryan:
Or, perhaps better, have a single property, such as `referenceDateType`, which you can set to an enum. That might be cleaner. What do you think?

ccgus:
An enum would be cleaner in my opinion.

kdubb:
After investigating this in-depth it looks like the issue is fixed by doing the math manually with a long double (aka 80bit); Cocoa even has a constant for the correct # of seconds with NSTimeIntervalSince1970.  Unfortunately SQLite doesn't support an 80bit real number.  After mulling the choices the decision is to compare NSDates that are stored in SQLite using the unix time value which has the proper reduced precision.

This actually negates the need for this change for our purposes.  I am happy change to an enum and prepare it for merging if it's still deemed useful in general.

PR #421, opened 2015-10-23
Updated README for Swift 2 changes

robertmryan:
The Swift example in README was referencing `println` and was using a deprecated `stringByAppendingPathComponent`. I've updated it for both of those issues.


ccgus:
Danke!

PR #425, opened 2015-10-30
Update documentation and class hierarchy

robertmryan:
Having issued a pull request for `executeQuery:values:error:` and `executeUpdate:values:error`, I updated class hierarchy and `index.html` accordingly.


robertmryan:
Related to https://github.com/ccgus/fmdb/pull/424

PR #426, opened 2015-11-03
Project file updates to support inclusion of FMDB via including the project in a workspace.

cutz:
This commit updates project settings to support building and referencing the FMDB and FMDB-IOS static libs by including the project in a workspace rather than via cocoa pods.  These two commits expose the base FMDB headers as public and adds a copy headers step such that headers can be included as <FMDB/*.h>.


ccgus:
Danke!

PR #434, opened 2015-11-24
Allow FTS to be compiled as Swift Framework

bdolman:
fts3_tokenizer.h only needs to be included internally by
FMDatabase+FTS3.m so we can exclude it from the umbrella header.
cc @newyankeecodeshop


bdolman:
Ah, hold off on this actually. It worked when I pointed to the local pod, not working when pointing to it remotely. I'll resubmit when it works.

PR #435, opened 2015-11-24
Allow FTS to be compiled as Swift framework

bdolman:
fts3_tokenizer.h only needs to be included privately by
FMDatabase+FTS3.m so we can exclude it from the umbrella header.
cc @newyankeecodeshop


ccgus:
I'm just assuming this works (I don't use pods)

newyankeecodeshop:
Looks good to me. I haven't started using Swift yet, but the FTS3 header definitely does not need to be exposed publicly.

PR #438, opened 2015-11-25
Update swift section of README to reflect experience with swift 7.2.1

peakman:
There is no file FMDB.m to add.  Any .m file seems to work.  The .m can be removed afterwards.

Creating a bridging header manually did not work for me as a required rule was then missing.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -267,9 +267,9 @@ You can use FMDB in Swift projects too.
 
 To do this, you must:
 
-1. Copy the FMDB `.m` and `.h` files into your project.
+1. Copy the FMDB.h file into your project and ANY .m file into your project.  The latter triggers a promprt for the necessary step of creating a bridging file.  It can be removed when it has done that.
 
-2. If prompted to create a "bridging header", you should do so. If not prompted and if you don't already have a bridging header, add one.
+2. Ccreate a "bridging header".  Creating one manually may not work as a required rule is not automatically generated.
 
  For more information on bridging headers, see [Swift and Objective-C in the Same Project](https://developer.apple.com/library/ios/documentation/Swift/Conceptual/BuildingCocoaApps/MixandMatch.html#//apple_ref/doc/uid/TP40014216-CH10-XID_76).
 


robertmryan:
I think this makes it more confusing than necessary. I see your confusion. I wasn't saying `FMDB.m` but rather all the `.m` and `.h` files in the FMDB source folder. I will clarify this.

But we don't want to advise adding random .m file to get the bridging header. If you add the `.m` files from the FMDB `src` folder, you'll be already be prompted to add a bridging header and that's sufficient. If you drag the whole folder over, that you won't be prompted. You have to drag the individual files. 

I will clarify.

peakman:
Hi Rob,

It’s not necessary to add all the files.  FMDB.h alone seems to be enough.

Unless I added a .m to the project (not to the bridging file) I wasn’t prompted to create a bridging file.  I agree that adding a random .m seems bizarre but it did work!  Creating a bridging file manually did NOT work.

Steve

peakman:
I’ve just re-read what you wrote and I understand better. But actually any .m will do so it’s not necessary to add them all.

Steve

robertmryan:
I understand what you're saying and I agree. I will take a crack at it.

robertmryan:
I have attempted to capture the spirit of this pull request in the latest commit, https://github.com/ccgus/fmdb/commit/3031658c74a587121c995234cdcbd3018855045a.

Thanks @peakman

peakman:
Hi Rob,

I read your updated README and it does make it clear which files can be copied.

However I think it really is more complicated now than it needs to be.

What I failed to mention earlier is that you do not need to include ANY of the .m files in order to use their contents.  They automatically get dragged in by the bridging header and FMDB.h.

It may well be nice to see the source  code that’s being used but it’s not necessary to have the .m files for a successful build.

Steve

robertmryan:
No, if you copy just `FMDB.h` into the project, that is not sufficient. When you go to use `FMDatabase` and the build your target, you'll get the following error:

    Undefined symbols for architecture x86_64:
      "_OBJC_CLASS_$_FMDatabase", referenced from:
          type metadata accessor for __ObjC.FMDatabase in ViewController.o
    ld: symbol(s) not found for architecture x86_64
    clang: error: linker command failed with exit code 1 (use -v to see invocation)

The only time you can get away with referencing `FMDB.h` only is if you built a separate target (e.g. a framework that included the `.m` files).

If there is a process by which you believe that you can add `FMDB.h` only and have it build targets successfully, please precisely outline the steps you're following (e.g. starting from creating new project). I cannot reproduce the behavior you describe. You _must_ include the `.m` files in the target (or build a framework that included the `.m` file).

As an aside, you reference "swift 7.2.1". Do you mean Xcode 7.1.1 running Swift 2.1?

peakman:
True, but as I said before you only need an arbitrary .m file to sort it out.

Simply copying FMDatabase.m is enough to give you access to all the .m files.

Steve

peakman:
You’re right.  It is Xcode 7.1.1 running Swift 2.1.  Sorry about that!

robertmryan:
What do you mean by "simply copying FMDatabase.m is enough to give you access to all the .m files"? That's simply not true. If you add `FMDatabase.m`, but not `FMResultSet.m`, to your project, when you compile, you'll get linking errors on `FMResultSet`:

    Undefined symbols for architecture x86_64:
      "_OBJC_CLASS_$_FMResultSet", referenced from:
          objc-class-ref in FMDatabase.o
    ld: symbol(s) not found for architecture x86_64
    clang: error: linker command failed with exit code 1 (use -v to see invocation)

peakman:
I think I’ve finally realised what’s going on!

I built a library from your project which I included in mine.  I did this when I couldn’t make things work in the first place because I misunderstood your instructions.  This is how the symbols  for FMResultSet etc are found.

I suppose I should have built a framework?

Anyway, at the end of this I think the easiest way for someone else is simply to copy all the .m files as you originally suggested and I misunderstood.

Sorry for wasting your time.  I’m a ruby person converting to swift so Xcode is a bit of a dark art still.

Steve

PR #450, opened 2015-12-21
Update README

robertmryan:
I've attempted to improve the "Data Sanitization" section, showing best practices in both Objective-C and Swift.

Before we used to have as many examples of what not to do as what to do, and I tried to, instead, just focus on what should be done. It makes it tighter (and avoids the reader who glances through it from accidentally grabbing one of the "bad" examples). I've also retired the `executeUpdateWithFormat` examples.

I've also adopted "Modern" (i.e. 3-4 years old) conventions of numeric and dictionary literals (e.g. `@42` is the same as `[NSNumber numberWithInt:42]` and `@{@"key" : @"value"}` instead of `[NSDictionary dictionaryWithObjectsAndKeys: @"value", @"key", nil]`. It just makes the examples a little more concise, IMHO.

But I wanted to issue this as a pull request so you could review, accept, improve, or reject, as you see fit.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -169,7 +169,7 @@ NSString *name = @"Liam O'Flaherty (\"the famous Irish author\")";
 NSDate *date = [NSDate date];
 NSString *comment = nil;
 
-BOOL success = [db executeUpdate:@"INSERT INTO myTable (identifier, name, date, comment) VALUES (?, ?, ?, ?)", @(identifier), name, date, comment ?: [NSNull null]];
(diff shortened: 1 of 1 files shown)


robertmryan:
BTW, reviewing edits in markup are annoying to read, so if you want to see it rendered, see https://github.com/robertmryan/fmdb/blob/master/README.markdown

ccgus:
Looks good- thanks for the fixes.

PR #454, opened 2015-12-29
Framework and Carthage Support

robertmryan:
Gus, here is version 2.6 pull request, which introduces framework support (akin to that of the `swiftFrameworks` branch, though with even more modest changes), as well as taking it the next step and introducing Carthage support (i.e. added two shared targets that build MacOS and iOS frameworks, respectively). I tried to update the change log to indicate the nature of the non-backward compatible change, but if you think we need something more explicit, let me know.

I haven't touched the Cocoapods stuff (other than bumping version), so perhaps we need someone more familiar with Cocoapods take a look at that.

I also haven't touched the static library targets (I'd vote for removing them to eliminate confusion, but that's your call). I'd also suggesting removing the command line target and `main.m`, too, because IMHO that's (a) confusing; and (b) is falling out of date).

On an unrelated topic, the Swift variadic extension for `FMDatabase` predates the `executeUpdate:values:error:` method that we added for Swift 2, so I've changed that it use the `values` rendition and throw errors. It's a non-backward compatible change (it changes this variadic extension to throw errors), but it's a good thing. Frankly, I would advise just removing it altogether, though, as I think the new `try db.executeUpdate("INSERT INTO ...", values ["foo", "bar"])` syntax is sufficiently concise that it eliminates the need for this variadic extension. And variadic methods are so un-Swift (though the language supports it). But, I know you love your variadic methods, so if you want to keep this tweaked extension, I'm fine with that.

### DIFF
--- a/CHANGES_AND_TODO_LIST.txt
+++ b/CHANGES_AND_TODO_LIST.txt
@@ -1,16 +1,23 @@
 TODO:
 Zip, nada, zilch.  Got any ideas?
 
-If you would like to contribute some code- awesome!  I just ask that you make it conform to the coding conventions already set in here, and to add a couple of tests for your new code to fmdb.m.  And of course, the code should be of general use to more than just a couple of folks.  Send your patches to gus@flyingmeat.com.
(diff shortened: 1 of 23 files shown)


ccgus:
Woo! Thanks for getting this all together.

robertmryan:
No worries. I also `tag`'ed this.

PR #466, opened 2016-02-04
Improve FTS tokenization performance

antons:
- `CFStringGetBytes` was calculating the UTF-8 offset from the beginning of string to current token again and again for each token. The longer the document the longer this operation took. Most tokenizers are sequential, and this repeated calculation was unnecessary
- Started caching last token range and its UTF-8 offset to avoid this repeated calculation.
- Extremely long documents (1 million characters) could take 30-40 seconds to index on a fast machine. Now they are indexed in less than a second
- Non-sequential tokenizers are still supported: cached range is reset if a token does not follow previous token.


ccgus:
Looks reasonable and tests pass. Thanks for this!

PR #469, opened 2016-02-22
Asynchronously perform database operations on queue, using transactions.

sunilsharma08:
synchronous transaction on queue hangs ui if some insertion/updation is performed on large no of rows.
To avoid ui hang because of synchronous transaction I have added a method named "inTransactionAsync" which perform database operations asynchronously.

### DIFF
--- a/src/fmdb/FMDatabaseQueue.h
+++ b/src/fmdb/FMDatabaseQueue.h
@@ -158,6 +158,13 @@
 
 - (void)inTransaction:(void (^)(FMDatabase *db, BOOL *rollback))block;
 
+/** Asynchronously perform database operations on queue, using transactions.
+ 
+ @param block The code to be run on the queue of `FMDatabaseQueue`
+ */
+
+- (void)inTransactionAsync:(void (^)(FMDatabase *db, BOOL *rollback))block;
+
 /** Synchronously perform database operations on queue, using deferred transactions.
 
  @param block The code to be run on the queue of `FMDatabaseQueue`
--- a/src/fmdb/FMDatabaseQueue.m
+++ b/src/fmdb/FMDatabaseQueue.m
@@ -210,6 +210,37 @@ static const void * const kDispatchQueueSpecificKey = &kDispatchQueueSpecificKey
     [self beginTransaction:NO withBlock:block];
 }
 
+- (void)inTransactionAsync:(void (^)(FMDatabase *db, BOOL *rollback))block{
+    [self beginTransactionAsync:NO withBlock:block];
+}
+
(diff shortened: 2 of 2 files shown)


ccgus:
What's the advantage of putting this in FMDB, vs just throwing it in a background block?

(PS- what's the point of the FMDBRetain & FMDBRelease outside the block?)

robertmryan:
FWIW, I second the vote that there should be an asynchronous rendition of `inTransaction` and `inDatabase` as part of FMDB. If a method can possibly block, it should be asynchronous. Apple has been religiously following this practice in all of their recent APIs and they just don't offer synchronous/blocking methods anymore. To be provocative, I think the question isn't "what's the advantage of putting this in FMDB?", but rather "should we have a synchronous rendition at all?"

My only question (setting aside the `FMDatabaseRetain` and `FMDatabaseRelease` issue) is whether this should be part of FMDB 2.x, or whether it should be deferred to FMDB 3.0. It seems to me that this might be better suited for the next major revision of FMDB. I say that because I think a better method name would be `performTransaction` or something like that. This name follows the verb-noun pattern that we use in APIs nowadays and avoids perpetuating the potential confusion of this `FMDatabaseQueue` method with the completely unrelated `inTransaction` method of `FMDatabase`. But such a major method renaming and/or shift from synchronous to asynchronous probably belongs in a major update, not a minor update.

ccgus:
I'm actually for an async method as well (as it's planned for 3.0: https://github.com/ccgus/fmdb/blob/three/FUTURE_PLANS.markdown )

I asked the question because I was curious.

3.0 has taken a backseat in my free time lately though. I had hoped to get something going last Dec, but it just didn't happen. Maybe this _should_ be part of 2.x, because I don't know when 3 will happen?

sunilsharma08:
> What's the advantage of putting this in FMDB, vs just throwing it in a background block?

My intention to add asynchronous method is to provide non blocking database operation even when working with large records.
I just tried to keep the same queue but operation should be performed asynchronously.

> (PS- what's the point of the FMDBRetain & FMDBRelease outside the block?)

I put the ```FMDBRetain``` & ```FMDBRelease``` inside the block in my second commit.

> "what's the advantage of putting this in FMDB?", but rather "should we have a synchronous rendition at all?"

It will be more appropriate to give both option to perform database operations either synchronously or asynchronously.Let the developer to choose which one is better for the task.

ccgus:
FWIW, I've just added async transactions to the three branch:
https://github.com/ccgus/fmdb/tree/three

Discussion on it here:
https://github.com/ccgus/fmdb/issues/487

PR #470, opened 2016-02-27
Fix mismatched-selectors warning

marcoarment:
Avoids the warning:
"Several methods with selector 'close' of mismatched types are found for the @selector expression"


ccgus:
Curious- what other class is implementing close differently?

ccgus:
Danke! I need to figure out why I've got both -close and -(BOOL)close in there, and maybe just use a single one in a future release.

PR #474, opened 2016-03-14
Map function for FMResultSet

natemann:
This is an implementation of a generic map function for FMResultSet. So instead of writing
```
let results = db.executeQuery(QUERY, values: nil)
results.next() {
    print(results.stringForColumn(COLUMN)
}
```

One can simply write
```
db.executeQuery(QUERY, values: nil).map { print($0.stringForColumn(COLUMN)) }
```

### DIFF
--- a/FMResultSet Extension.swift
+++ b/FMResultSet Extension.swift
@@ -1,0 +1,18 @@
+//
+//  FMResultSet Extension.swift
+//
+
+import Foundation
+
+extension FMResultSet {
+
+    ///A map function to replace `resultSet.next()`
+    func map<T>(transform: FMResultSet -> T) -> [T] {
+        var results: [T] = []
+        while self.next() {
+            results.append(transform(self))
+        }
+        return results
+    }
+    
+}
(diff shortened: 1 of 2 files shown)


ccgus:
@robertmryan - how does this guy look to you? I'm still not very comfy with Swift yet.

robertmryan:
With no offense intended, I'm not crazy about this `map` implementation. It strikes me as a fundamental misunderstanding the `map` function, viewing it as merely syntactic sugar for a `for` loop that returns an array. But `map` is designed to take some stateless collection type (usually an array) and build a new collection from it. 

```swift
let numbers = [20, 19, 7, 12]
let triples = numbers.map { $0 * 3 }  // now [60, 57, 21, 36]
```

Sure, `map` behaves a little like a `for` loop, but its roots rest in Haskell, and it's intended for functional programming. The `map` function takes some collection (and, FWIW, `FMResultSet` is not a collection type), performing some action on each member of that collection, and returning another collection consisting of the elements of the original collection, each of which have been transformed in some way. And, when doing functional programming, the `transform` closure (the `$0 * 3` in my example) really should not have any side-effects (e.g. shouldn’t be printing anything, doing anything with external variables, etc., but rather just performing a `transform` on each member of the original collection in order to return a new collection). 

But `FMResultSet` is not a stateless collection type. It has an internal state (i.e. how far you've iterated through the result set) and calling this `map` function is changing its state. This is antithetical to functional programming paradigm. And because SQLite doesn't have a "seek" like function (where you can reset the state of the `sqlite3_stmt` to return where you were before), I'm having a hard time mentally reconciling `sqlite3_stmt` (which rests at the heart of `FMResultSet`) with good functional programming practices.

The only thing that seems a bit more "Swifty" would be to have a rendition of `executeQuery` that returned a collection instead of a `FMResultSet` (e.g. perhaps a `resultDictionaryArray` method on `FMDatabase` that returned returned an array of `resultDictionary` objects). That would be something that could dovetail into Swift's existing `map` function (though it would never be up to Swift's strong typing, thanks to SQLite's pusillanimous notion of data types).

In short, I'd suggest we table this for now and think about it. But I'd lean towards something that returned an array of `resultDictionary` (or some new `Result` type), which could then use existing Swift functional programming patterns for free.

natemann:
@robertmryan, could we perhaps rename this function? Instead of `map`, maybe `next`, or `resultsNext`?  This would take away any confusion with the history of `map` functions.

robertmryan:
While I appreciate the spirit of this pull request, I am not inclined to incorporate it into the master branch of FMDB at this point. But you've given us food for thought for features in future versions of FMDB. Thanks!

Obviously, just because we don't fold this into FMDB, itself, there's no reason why you can't use this extension in your own projects. That's the wonderful thing about extensions, that you can add whatever extensions you want, whether to Cocoa classes or FMDB classes.

PR #480, opened 2016-04-05
Added support for custom database class and vfsName

gloubibou:
FMDatabasePool can now be subclassed to use a subclass of FMDatabase.
FMDatabasePool initializer now accepts a custom vfsName.

### DIFF
--- a/src/fmdb/FMDatabasePool.h
+++ b/src/fmdb/FMDatabasePool.h
@@ -39,7 +39,8 @@
     __unsafe_unretained id _delegate;
     
     NSUInteger          _maximumNumberOfDatabasesToCreate;
-    int                 _openFlags;
+	int                 _openFlags;
+	NSString            *_vfsName;
 }
 
 /** Database path */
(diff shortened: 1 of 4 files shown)


ccgus:
Looks OK to me (though, tabs are being used instead of spaces- care to fix that?).

@robertmryan - any comments on this?

robertmryan:
I guess it's OK, though it doesn't feel 100% right. It feels like there's some candidate for generics (though Objective-C only has "lightweight generics" at this point) or something like. And maybe there should be some assertion (not `NSAssert`; but production fatal error) to make sure that `databaseClass`, in fact, instantiates a `FMDatabase` subclass.

I might be inclined to keep `databaseClass` a private method (and go ahead and let gloubibou carry on with his subclassing approach for now), but let us stew over better, type-safe ways to do this in the long run.

gloubibou:
I don’t actually use `databaseClass`. I added it to FMDatabasePool since it already existed on FMDatabaseQueue. I do use vfsName.

gloubibou:
@ccgus I replaced the tabs with space

ccgus:
Wait, so you added code that's never been executed? Boo! :)

It looks reasonable though, and it's private, so we'll keep it. Thanks for the patches!

PR #483, opened 2016-04-27
Use singleton pattern to rewrite FMDB.h

ZXIOU:
Hi~, I'm ZXIOU, I rewrite this because database is a shared resource,  I want to change the FMDB
class to a singleton class to simplify the use of FMDB


ccgus:
Thanks for the pull request, but this isn't something we want to add to FMDB.

ZXIOU:
Ok, I will learn more to contribution to FMDB to make it better.

PR #486, opened 2016-05-10
Set APPLICATION_EXTENSION_API_ONLY to YES

anagromataf:
Allow using FMDB in an application extension by setting the flag APPLICATION_EXTENSION_API_ONLY to YES.

### DIFF
--- a/fmdb.xcodeproj/project.pbxproj
+++ b/fmdb.xcodeproj/project.pbxproj
@@ -885,6 +885,7 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				ALWAYS_SEARCH_USER_PATHS = NO;
+				APPLICATION_EXTENSION_API_ONLY = YES;
 				CLANG_CXX_LANGUAGE_STANDARD = "gnu++0x";
 				CLANG_CXX_LIBRARY = "libc++";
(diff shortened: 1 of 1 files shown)


ccgus:
Why would FMDB need this? Is this a thing for external libraries (and if that's the case- wouldn't it only be needed on two or three targets?)

anagromataf:
@ccgus This makes sure, that FMDB only uses API that is allowed in extensions. For example, I'm using FMDB in a framework, that could be used in an app extension. Without this flag I always get those warnings. This for sure only applies if you are using frameworks (e.g., via carthage) and not just the sources.

benshan:
We have the same issue and locally modify fmdb with this change. It would be great to merge this since the library already does not use forbidden methods and this change will prevent such breaking changes in the future, along with silencing the warning when fmdb is linked to an extension.

benshan:
@ccgus do you need any other info to move this forward? Without this change, consumers cannot safely link to the iOS FMDB framework from an iOS extension or from a custom framework which is marked as supporting extensions. Given Apple's focus on expanding integrations into iOS via extension, this has become more of an issue.

ccgus:
It's merged.

benshan:
Thanks @ccgus, much appreciated!

Issue #490, opened 2016-05-17
cocoapods 1.0 default_subspecs


donileo:
I am getting an error installing fmdb with cocoapods 1.0

[!] The `FMDB` pod failed to validate due to 1 error:
    - ERROR | attributes: Can't set `default_subspecs` attribute for subspecs (in `FMDB/standalone`).

srivkas:
I have this issue too, this happen because only the root spec can have "default subspecs" please delete default_subspecs not in the root element. what can i do till you will fix this issue?

ccgus:
Someone will need to submit a patch for this.

ahti:
I think just deleting the default_subspec for the standalone subspect would cause that subspec to always be built with FTS support, which I'm not sure is desirable. I'm also not very experienced in writing podspecs so I may be wrong.

@blakewatters, you authored the last few significant commits to the podspec. Do you know what the right solution here is?

BennettSmith:
Submitted pull request that resolves this issue.  See https://github.com/ccgus/fmdb/pull/496

ccgus:
I've merged it- thanks @BennettSmith .

PR #496, opened 2016-05-24
Resolves default_subspecs error in podfile.

BennettSmith:
It looks like Cocoapods is not more restrictive about how default_subspecs is used.  In order to continue supporting alternative builds with the latest sqlite3 and FTS this pull request gets rid of the nested subspecs.  Instead of 'standalone' with 'default' and 'FTS' nested within, the pod spec now contains two alternative subspecs for 'standalone' and 'standalone-fts'.


BennettSmith:
Resolves issue #490

PR #499, opened 2016-06-07
Do not pass nil to va_list arguments.

sas:
Some systems/compilers do not define va_list as being a pointer type.

### DIFF
--- a/src/fmdb/FMDatabase.m
+++ b/src/fmdb/FMDatabase.m
@@ -10,8 +10,8 @@
 
 @interface FMDatabase ()
 
-- (FMResultSet *)executeQuery:(NSString *)sql withArgumentsInArray:(NSArray*)arrayArgs orDictionary:(NSDictionary *)dictionaryArgs orVAList:(va_list)args;
-- (BOOL)executeUpdate:(NSString*)sql error:(NSError**)outErr withArgumentsInArray:(NSArray*)arrayArgs orDictionary:(NSDictionary *)dictionaryArgs orVAList:(va_list)args;
+- (FMResultSet *)executeQuery:(NSString *)sql withArgumentsInArray:(NSArray*)arrayArgs orDictionary:(NSDictionary *)dictionaryArgs orVAList:(va_list*)args;
+- (BOOL)executeUpdate:(NSString*)sql error:(NSError**)outErr withArgumentsInArray:(NSArray*)arrayArgs orDictionary:(NSDictionary *)dictionaryArgs orVAList:(va_list*)args;
 
 @end
 
@@ -744,7 +744,7 @@ static int FMDBDatabaseBusyHandler(void *f, int count) {
     return [self executeQuery:sql withArgumentsInArray:nil orDictionary:arguments orVAList:nil];
 }
 
-- (FMResultSet *)executeQuery:(NSString *)sql withArgumentsInArray:(NSArray*)arrayArgs orDictionary:(NSDictionary *)dictionaryArgs orVAList:(va_list)args {
+- (FMResultSet *)executeQuery:(NSString *)sql withArgumentsInArray:(NSArray*)arrayArgs orDictionary:(NSDictionary *)dictionaryArgs orVAList:(va_list*)args {
     
     if (![self databaseExists]) {
         return 0x00;
@@ -834,7 +834,7 @@ static int FMDBDatabaseBusyHandler(void *f, int count) {
                 obj = [arrayArgs objectAtIndex:(NSUInteger)idx];
             }
             else if (args) {
-                obj = va_arg(args, id);
+                obj = va_arg(*args, id);
             }
             else {
                 //We ran out of arguments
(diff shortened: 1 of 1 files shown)


ccgus:
Can you give an example of a system or compiler where this is an issue?

sas:
(Sorry for the delay, I missed this email.)

We're basically at a phase where we're experimenting building this code on
non-Apple platforms, so we're trying to limit how much the code we use
relies on behavior that is not guaranteed by the standard. In this case,
the C standard does not say anything about `va_list` being a pointer type;
it only says a pointer to a va_list can be taken and passed to other
functions (which is what this code does).

Let me know if this is something you think is acceptable for your project.

Thanks :)

On Tue, Jun 7, 2016 at 10:20 AM, ccgus <notifications@github.com> wrote:

> Can you give an example of a system or compiler where this is an issue?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/ccgus/fmdb/pull/499#issuecomment-224351004>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AADHbFknDry-YmL1YjBbOyqCOQp21TtFks5qJahNgaJpZM4IveXu>
> .
>



-- 
Stephane Sezer

ccgus:
I'm revisiting this one right now, and I don't see anything inherently wrong with it- @robertmryan do you have any comments on this patch?

I breaks a signature, but it's a internal interface so it shouldn't matter.

robertmryan:
I've generally seen `va_list`, not a pointer to `va_list`. E.g. http://c-faq.com/varargs/handoff.html

In this discussion, http://stackoverflow.com/a/3369762/1271826, there is a quote from page 249 of the [C99 standard](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1124.pdf) which suggests `va_list` is an opaque type that can safely be passed around:

> The object `ap` may be passed as an argument to another function; if that function invokes the `va_arg` macro with parameter `ap`, the value of `ap` in the calling function is indeterminate and shall be passed to the `va_end` macro prior to any further reference to `ap`.

But there is a footnote that suggests there are special cases where one can (but not necessarily should) pass a pointer:

> It is permitted to create a pointer to a `va_list` and pass that pointer to another function, in which case the original function may make further use of the original list after the other function returns.

So my reading of all of this is that one can certainly pass a `va_list`, but that you could pass a pointer if the original function was going to "make further use" of it, which it isn't here.

So, my bottom line is that I'm not following the case for changing it to a pointer here.

appleguy:
@sas @robertmryan The issue is that a NULL va_list is not valid, but a NULL pointer to a va_list is valid.  va_list is similar to a struct type in this sense.

We are using FMDB at Pinterest and also have an issue with this with our compiler and static analysis, and using this patch addresses it.  As long as it feels safe, I'm pretty confident this is more formally correct, and we'd be happy to eliminate our (very simple) fork of FMDB by integrating this..

robertmryan:
BTW, when I replace my `FMDatabase.m` with this, when I compile this in Xcode Version 7.3.1 (7D1014), I get warnings:

<pre>
/Users/rryan/.../FMDatabase.m:929:86: warning: incompatible pointer types sending 'struct __va_list_tag **' to parameter of type 'va_list *' (aka '__builtin_va_list *') [-Wincompatible-pointer-types]
    return [self executeQuery:sql withArgumentsInArray:nil orDictionary:nil orVAList:&args];
                                                                                     ^~~~~
/Users/rryan/.../FMDatabase.m:13:151: note: passing argument to parameter 'args' here
- (FMResultSet *)executeQuery:(NSString *)sql withArgumentsInArray:(NSArray*)arrayArgs orDictionary:(NSDictionary *)dictionaryArgs orVAList:(va_list*)args;
                                                                                                                                                      ^
/Users/rryan/.../FMDatabase.m:1175:97: warning: incompatible pointer types sending 'struct __va_list_tag **' to parameter of type 'va_list *' (aka '__builtin_va_list *') [-Wincompatible-pointer-types]
    return [self executeUpdate:sql error:nil withArgumentsInArray:nil orDictionary:nil orVAList:&args];
                                                                                                ^~~~~
/Users/rryan/.../FMDatabase.m:14:166: note: passing argument to parameter 'args' here
- (BOOL)executeUpdate:(NSString*)sql error:(NSError**)outErr withArgumentsInArray:(NSArray*)arrayArgs orDictionary:(NSDictionary *)dictionaryArgs orVAList:(va_list*)args;
                                                                                                                                                                     ^
2 warnings generated.
</pre>

ccgus:
The tests pass with the patch and logically things make sense to me, but the warnings really concern me. va_list seems like it contains quite a bit of compiler black magic, and I guess I'm not comfortable with the change.

I can't get the static analyzer to complain about anything in fmdb right now- I guess since you're using another platform that probably means a non llvm compiler as well. Is there any other way to approach this so fmdb doesn't generate warnings and everything is safe on darwin?

PR #502, opened 2016-06-07
fix issues with bool binding

RishatShamsutdinov:


RishatShamsutdinov:
How to reproduce issue:
`BOOL b = 0x10`.

Use value from `b` to update db. You will see `16` instead of `1`. I use indexes so I have to use condition like `unread = 1` instead of `unread != 0`.

Reason of issue: `strcmp([obj objCType], @encode(char)) == 0` returns `true` for BOOL values.

Environment: iPhone 4S (iOS 8/9). On iPhone 5 & newer you will not reproduce this issue.

robertmryan:
This looks promising. Could we impose upon you to add a test case for this? Either supplement the existing `testCharAndBoolTypes` test case, or perhaps better, add your own test case?

ccgus:
The char bool strikes again. I've never liked that, not sure why it couldn't have been 1 bit.

robertmryan:
Yeah, it's the old C heritage of the language. Fortunately, on newer platforms it now is true Boolean. But I think this pull request is trying to make the behavior more consistent on older platforms. And this is, admittedly addressing the edge case where someone incorrectly uses value other than `YES`/`NO` or `TRUE`/`FALSE` with `BOOL` on these older platforms.

RishatShamsutdinov:
"Incorrectly" usage can be caused by the following code: `BOOL unread = (flags & UNREAD_FLAG)` where `UNREAD_FLAG = 1 << 3`, for example.

@robertmryan I'll create new PR with tests to compare version with this PR & w/o this PR.

RishatShamsutdinov:
My bad. `@((BOOL)0x10)` always produce `kCFBooleanTrue`.

PR #401, opened 2015-08-28
Add FMDB/SQLCipher/FTS subspec

evands:
This allows SQLCipher+FTS.

PR #503, opened 2016-06-08
Replacement for #401: Update podspec for SQLCipher+fts3

evands:
It turned out to be significantly easier and clearer to just create a new branch and pull request for this fix. The first commit just simplifies the podspec to use dependencies to deduplicate code; the second adds FMDB/SQLCipher/FTS3

### DIFF
--- a/FMDB.podspec
+++ b/FMDB.podspec
@@ -7,7 +7,12 @@ Pod::Spec.new do |s|
   s.author = { 'August Mueller' => 'gus@flyingmeat.com' }
   s.source = { :git => 'https://github.com/ccgus/fmdb.git', :tag => "#{s.version}" }
   s.requires_arc = true
-  s.default_subspec = 'standard'  
+  s.default_subspec = 'standard'
+
+  s.subspec 'common' do |ss|
+    ss.source_files = 'src/fmdb/FM*.{h,m}'
+    ss.exclude_files = 'src/fmdb.m'
+  end
 
   # use the built-in library version of sqlite3
   s.subspec 'standard' do |ss|
@@ -26,24 +31,21 @@ Pod::Spec.new do |s|
   s.subspec 'standalone' do |ss|
     ss.xcconfig = { 'OTHER_CFLAGS' => '$(inherited) -DFMDB_SQLITE_STANDALONE' }
     ss.dependency 'sqlite3'
-    ss.source_files = 'src/fmdb/FM*.{h,m}'
-    ss.exclude_files = 'src/fmdb.m'
-  end
-
-  # build with FTS support and custom FTS tokenizer source files
+    ss.dependency 'FMDB/common'
+	end
+	
+  # build the latest stable version of sqlite3 
+	# with FTS support and custom FTS tokenizer source files
   s.subspec 'standalone-fts' do |ss|
-    ss.xcconfig = { 'OTHER_CFLAGS' => '$(inherited) -DFMDB_SQLITE_STANDALONE' }
-    ss.source_files = 'src/fmdb/FM*.{h,m}', 'src/extra/fts3/*.{h,m}'
-    ss.exclude_files = 'src/fmdb.m'
+    ss.dependency 'FMDB/standalone'
     ss.dependency 'sqlite3/fts'
+    ss.source_files = 'src/extra/fts3/*.{h,m}'
   end
 
   # use SQLCipher and enable -DSQLITE_HAS_CODEC flag
   s.subspec 'SQLCipher' do |ss|
     ss.dependency 'SQLCipher'
-    ss.source_files = 'src/fmdb/FM*.{h,m}'
-    ss.exclude_files = 'src/fmdb.m'
-    ss.xcconfig = { 'OTHER_CFLAGS' => '$(inherited) -DSQLITE_HAS_CODEC -DHAVE_USLEEP=1' }
-  end
-  
+  	ss.dependency 'FMDB/common'
+    ss.xcconfig = { 'OTHER_CFLAGS' => '$(inherited) -DSQLITE_HAS_CODEC -DHAVE_USLEEP=1' }		
+  end  
 end
--- a/FMDB.podspec
+++ b/FMDB.podspec
@@ -47,5 +47,12 @@ Pod::Spec.new do |s|
     ss.dependency 'SQLCipher'
   	ss.dependency 'FMDB/common'
     ss.xcconfig = { 'OTHER_CFLAGS' => '$(inherited) -DSQLITE_HAS_CODEC -DHAVE_USLEEP=1' }		
+		
+		# FMDB/SQLCipher/FTS subspec, which uses SQLCipher 
+		# and adds FTS support + custom FTS tokenizer source files
+    ss.subspec 'FTS' do |sss|
+        sss.dependency 'SQLCipher/fts'
+        sss.dependency 'FMDB/FTS'
+    end
   end  
 end


ccgus:
@stephanheilner - can you quickly review this (since you seem to use cocoa pods) and let me know if it looks OK?

FMDB.podspec
```diff
@@ -26,24 +31,28 @@ Pod::Spec.new do |s|
...
 
   # use SQLCipher and enable -DSQLITE_HAS_CODEC flag
   s.subspec 'SQLCipher' do |ss|
     ss.dependency 'SQLCipher'
-    ss.source_files = 'src/fmdb/FM*.{h,m}'
-    ss.exclude_files = 'src/fmdb.m'
-    ss.xcconfig = { 'OTHER_CFLAGS' => '$(inherited) -DSQLITE_HAS_CODEC -DHAVE_USLEEP=1' }
-  end
-  
+  	ss.dependency 'FMDB/common'
```
stephanheilner: Though it wouldn't affect how this functions, the indentation on this is wrong. it should like up with:
```
ss.dependency 'SQLCipher'
ss.dependency 'FMDB/common'
ss.xcconfig = ...
```

FMDB.podspec
```diff
@@ -26,24 +31,28 @@ Pod::Spec.new do |s|
   s.subspec 'standalone' do |ss|
     ss.xcconfig = { 'OTHER_CFLAGS' => '$(inherited) -DFMDB_SQLITE_STANDALONE' }
     ss.dependency 'sqlite3'
-    ss.source_files = 'src/fmdb/FM*.{h,m}'
-    ss.exclude_files = 'src/fmdb.m'
-  end
-
-  # build with FTS support and custom FTS tokenizer source files
+    ss.dependency 'FMDB/common'
+	end
```
stephanheilner: Indentation on `end` is wrong here

FMDB.podspec
```diff
@@ -26,24 +31,28 @@ Pod::Spec.new do |s|
...
+	
+  # build the latest stable version of sqlite3
```
stephanheilner: I'd tweak this comment to something like: 
`# Use the standalone version of sqlite3/fts and include custom FTS Tokenizer code`

FMDB.podspec
```diff
@@ -7,7 +7,12 @@ Pod::Spec.new do |s|
...
-  s.default_subspec = 'standard'  
+  s.default_subspec = 'standard'
+
+  s.subspec 'common' do |ss|
+    ss.source_files = 'src/fmdb/FM*.{h,m}'
+    ss.exclude_files = 'src/fmdb.m'
+  end
 
   # use the built-in library version of sqlite3
   s.subspec 'standard' do |ss|
```
stephanheilner: `s.subspec 'standard' do |ss|` should just have the `ss.dependency 'FMDB/common'` dependency instead of duplicating the source_files and exclude_files.

FMDB.podspec
```diff
@@ -26,24 +31,28 @@ Pod::Spec.new do |s|
...
-    ss.exclude_files = 'src/fmdb.m'
-  end
-
-  # build with FTS support and custom FTS tokenizer source files
+    ss.dependency 'FMDB/common'
+	end
+	
+  # build the latest stable version of sqlite3 
+	# with FTS support and custom FTS tokenizer source files
   s.subspec 'standalone-fts' do |ss|
```
stephanheilner: I *think* that you could add a sub-subspec to `s.subspec 'standalone' do |ss|` for FTS. Something like this:

```
s.subspec 'standalone' do |ss|
    ss.xcconfig = { 'OTHER_CFLAGS' => '$(inherited) -DFMDB_SQLITE_STANDALONE' }
    ss.dependency 'sqlite3'
    ss.dependency 'FMDB/common'
    ss.subspec 'FTS' do 'sss'
        sss.dependency 'sqlite3/fts'
        sss.source_files = 'src/extra/fts3/*.{h,m}'
    end
end
```
You'll have to verify though...

FMDB.podspec
```diff
@@ -26,24 +31,28 @@ Pod::Spec.new do |s|
...
     ss.dependency 'SQLCipher'
-    ss.source_files = 'src/fmdb/FM*.{h,m}'
-    ss.exclude_files = 'src/fmdb.m'
-    ss.xcconfig = { 'OTHER_CFLAGS' => '$(inherited) -DSQLITE_HAS_CODEC -DHAVE_USLEEP=1' }
-  end
-  
+  	ss.dependency 'FMDB/common'
+    ss.xcconfig = { 'OTHER_CFLAGS' => '$(inherited) -DSQLITE_HAS_CODEC -DHAVE_USLEEP=1' }		
+		
+		# FMDB/SQLCipher/FTS subspec, which uses SQLCipher
```
stephanheilner: These comments need to be indented correctly

FMDB.podspec
stephanheilner: I would add the subspec for FTS here as well:

```s.subspec 'standard' do |ss|
    ss.library = 'sqlite3'
    ss.source_files = 'src/fmdb/FM*.{h,m}'
    ss.exclude_files = 'src/fmdb.m'

    # use the built-in library version of sqlite3 with custom FTS tokenizer source files
    ss.subspec 'FTS' do |sss|
        sss.source_files = 'src/extra/fts3/*.{h,m}'
        sss.pod_target_xcconfig = { 'OTHER_CFLAGS' => '$(inherited) -DSQLITE_ENABLE_FTS4=1 -DSQLITE_ENABLE_FTS3_PARENTHESIS=1' }
    end
end
```

stephanheilner:
I added some suggestions and made some comments inline.

FMDB.podspec
evands: Agreed - that was an oversight. Thanks for catching it.

FMDB.podspec
evands: Agreed that this is better.  I'd just kept the separate subspec to minimize changes from the existing podfile.

FMDB.podspec
evands: I like this idea and it's similar to how it was being done a long time ago in this podfile.  As above, I left the weirdness to minimize the changes I was making. Better to just fix it properly.

evands:
@stephanheilner: Please see [300d0fe] and let me know if you have any further suggestions.  I think we've got a much cleaner podspec now as a result of these efforts.

evands:
Finally, we'll need to bump the podfile version number and have a corresponding git tag, once this is approved.

ccgus:
So everyone is happy with this now? All I need to do is commit it?

stephanheilner:
We'll need to update the version from 2.6.2 to whatever the next tag version will be, but we can do that right before @ccgus creates a new tag. Just give me a heads up, so I can make sure you bump the version in this file as well first.

evands:
I'm having difficulty with custom tokenizers working in SQLCipher+FTS.  Please don't commit yet -- investigating.

evands:
OK, I figured out the problem but don't have an immediate solution.  @stephanheilner would love your input.  
 * SQLCipher needs to be compiled with -DDSQLITE_ENABLE_FTS3_TOKENIZER for custom tokenizers to work.  Adding this in our podspec doesn't change how the dependency is compiled so doesn't fix the problem.
 * Maybe should be done in [the podspec file](https://github.com/CocoaPods/Specs/blob/master/Specs/SQLCipher/3.4.0/SQLCipher.podspec.json) I think but I can't find a source podspec in [their github repo](https://github.com/sqlcipher/sqlcipher ) to make a pull request

pleasantnicholasb:
Just wondering if this branch is still being worked on.

evands:
I concluded it probably needs to be fixed at the level of SQLCipher if it is to be properly done and couldn't find anywhere to do that.  I've just added this to my Podfile:
```
post_install do |installer|
 puts "Post-Install: Enabling SQLITE_ENABLE_FTS3_TOKENIZER for SQLCipher..."	
  sqlcipher_targets = installer.pods_project.targets.select { |target| target.name =~ /^SQLCipher$/ }
  sqlcipher_targets.each do |target|
        target.build_configurations.each do |config|
            old_defines = config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] || []
            config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] = old_defines + ['$(inherited) SQLITE_ENABLE_FTS3_TOKENIZER=1']
        end
    end
end
```

RishatShamsutdinov:
`-DDSQLITE_ENABLE_FTS3_TOKENIZER` is not required because you can enable custom tokenizers by following code:
`sqlite3_db_config(sqliteHandle, SQLITE_DBCONFIG_ENABLE_FTS3_TOKENIZER, 1, 0);`

PR #504, opened 2016-06-08
Re-add the ability to interrupt a query in progress

evands:
Marek Serafin's [cbc4188] was previously applied, but the changes were silently removed in [59f8d63]. This reapplies those changes. 

(Theoretically this should be a cherrypick, but I ended up in an hourlong fight with git over it which I think are related to whitespace changes between the 2014 commit and now).

### DIFF
--- a/src/fmdb/FMDatabase.h
+++ b/src/fmdb/FMDatabase.h
@@ -734,6 +734,15 @@ typedef int(^FMDBExecuteStatementsCallbackBlock)(NSDictionary *resultsDictionary
 
 - (void)setShouldCacheStatements:(BOOL)value;
 
+/** Interupt pending database operation
+ 
+ This method causes any pending database operation to abort and return at its earliest opportunity
+ 
+ @return `YES` on success; `NO` on failure. If failed, you can call `<lastError>`, `<lastErrorCode>`, or `<lastErrorMessage>` for diagnostic information regarding the failure.
+ 
+ */
+
+- (BOOL)interrupt;
 
 ///-------------------------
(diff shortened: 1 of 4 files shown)


ccgus:
Thanks for this! A couple of questions before I commit this:

1) How would you test this?
2) Why wouldn't closing a result set do the same thing? How is this better / different / more advantageous?

evands:
Some background: 
In the app where I make use of interrupt, I have a search field which triggers a full-text search of a large database.  That search is ultimately performed via a single `FMDatabaseQueue`.  However, a full-text search can be a non-trivial query, especially when searching for a short string.  

When the search text changes, dispatch_async() is used to move the search process to a new background thread. -[FMDatabaseQueue inDatabase:] will be used on that thread (so that appropriate sequential use of a single thread for the database  occurs).  However, if the search text has changed, the old query is completely irrelevant and we want to make it stop immediately in favor of the new query. Calling interrupt allows this to happen.  It isn't possible to close the result set because it doesn't exist until the query is finished.

(For example, the user is typing 'hello'.  A query began at 'h', is canceled in favor of 'he', which is canceled in favor of 'hel', and so forth.)

evands:
I'm not sure how best to test it systematically - I guess a sufficiently time-consuming query would be needed.

ccgus:
Huh- I had no idea it would stop a query before the result sets are returned- that's awesome. I could have used that back when I was working on VoodooPad.

Looks like it's cool to call it from another thread too, which was another concern of mine. Danke!

PR #523, opened 2016-08-29
Fix a crash if value isn't UTF8 string

elliotfiske:


ccgus:
Hello @elliotfiske - can you make a unit test that shows the crash in action?

elliotfiske:
updated! Let me know if you have any questions/I did something wrong.

ccgus:
In your first patch, you have it setting the value to null, and then in another one you have it set to [NSString stringWithCString:values[i] encoding:NSASCIIStringEncoding].

Falling back on NSASCIIStringEncoding is a bad idea in my opinion, because we don't do that anywhere else, and UTF-8 is the way to go anyway.

Any reason why you changed it from NSNull?

elliotfiske:
I changed it from NSNull because there _is_ actually data there, it's just not encodable to a UTF-8 string.

I'm actually thinking it might make more sense to return NSData now, do you think that would make more sense?

ccgus:
I think putting anything other than NSString at this point would be a bad idea, because that could break applications today. However, for 3.0 it would make sense to look at the sqlite type being returned, and then massage the data to that.

So for now, it should probably be NSNull (which is what FMDB does elsewhere, so it's consistent at least)

elliotfiske:
I think for now we should keep the noisy crash instead of silently returning Null for BLOB data, and push off the problem off to 3.0, especially since I realize now we might return bad data – if the binary data has a 0-byte, it'll truncate it because it's a C-style string.

PR #529, opened 2016-09-07
Orm additions

tblanchard:
Some helper categories.  I will understand if you don't pull them.  I make use of them in a little active record implementation I have.  I will contribute that too once I get it cleaned up and separated out of my current project.


ccgus:
ORM is a bit beyond the scope of FMDB, but thanks for trying to contribute, we appreciate it even if we don't take everything :)

Issue #535, opened 2016-09-19
Unknown error calling sqlite3_step (10: disk I/O error) rs


tedrog36:
Just started getting error or at least getting it more often on iOS10.  

See this link for more info: https://github.com/marcoarment/FCModel/issues/105

Any ideas of what might be happening?  The code is currently built with FMDB 2.5.

Thanks.

ccgus:
Show all the backtraces please.

tedrog36:
In this stacktrace thread 23 is sitting on this statement in nextWithError:
        // wtf?
        NSLog(@"Unknown error calling sqlite3_step (%d: %s) rs", rc, sqlite3_errmsg([_parentDB sqliteHandle]));

[StackTrace.txt](https://github.com/ccgus/fmdb/files/481002/StackTrace.txt)

ccgus:
I haven't looked at FCModel, so I can only guess as to what's going on, but it looks like both thread 22 & 23 are doing database operations at the same time, which is a bad idea.

tedrog36:
That is what I thought at first but thread 22 and 23 use the same operation queue and it appears that 22 is waiting for 23 to finish.

tedrog36:
Gus - I am just learning this codebase I am debugging.  I realize that it is writing images to the database as JPEG NSData.

Any experience with issues around doing this? 

My current hypothesis is that writing these large images to the database is driving this problem.

ccgus:
Acorn uses FMDB, and writes some very large images to databases (100mp even), so I don't think that's it. I think there's a threading error somewhere.

tedrog36:
I update to the latest FCModel which puts all database requests on the main UI thread so I don't think it is that.

ccgus:
You still get the problem when all requests are on the main thread? Can you show all the backtraces when that happens?

tedrog36:
See attached stacktrace from Crashlytics.
[crashlytics_stacktrace.txt](https://github.com/ccgus/fmdb/files/485668/crashlytics_stacktrace.txt)

ccgus:
This isn't the same error as before. The app is dying because of a thrown exception.

tedrog36:
Yes.  "Query failed with SQLite error 10: disk I/O error"

Which is the same root cause as in the previous stack trace.

ccgus:
I don't know what's causing, sorry. A reproducible case (in source code form) would help.

tedrog36:
[...]

PR #536, opened 2016-09-22
Expose `sqlite3_extended_errcode`

robertmryan:
Pursuant to issue https://github.com/ccgus/fmdb/issues/535, we may want to expose `sqlite3_extended_errcode`.

Also, unrelated, updated notes for variadic `FMDatabase` methods, to remove the advise to add the variadic extension. Better to just use `executeQuery:values:`, IMHO.


ccgus:
Danke!

Issue #538, opened 2016-09-30
Crash "freeing unallocated pointer" in FMDatabase.m line 1073


georgbachmann:
My app wasn't yet recompiled for iOS10, but since iOS10 got released, I have a lot of those crashes:
    
    Crashed: fmdb.<FMDatabaseQueue: 0x176653c50>
    SIGABRT ABORT 0x0000000188292014

in `FMDatabase.m line 1073`I am using Version 2.6.2

The main problem is that I can't reproduce the crash...
Also the stacktrace isn't very helpful to me... 

    FMDatabase.m line 1073
    -[FMDatabase executeUpdate:error:withArgumentsInArray:orDictionary:orVAList:]
    
    FMDatabase.m line 1239
    -[FMDatabase executeUpdate:withErrorAndBindings:]
    
    BFTourInfoDatabase.m line 377
    __45-[BFTourInfoDatabase updateTours:completion:]_block_invoke_2
    
    FMDatabaseQueue.m line 192
    __46-[FMDatabaseQueue beginTransaction:withBlock:]_block_invoke
    
    _dispatch_client_callout + 16
    
    _dispatch_barrier_sync_f_invoke + 84
    
    FMDatabaseQueue.m line 203
    -[FMDatabaseQueue beginTransaction:withBlock:]

Does anybody else have this crash? Any help is very appreciated as I'd like to update my app soon to fix this problem!

roblav96:
@georgbachmann Same here. Have you found a solution?

ccgus:
Please show the full backtrace, and code snippets that are calling FMDB would be helpful as well.

georgbachmann:
I don't have more than the snipped above. I also can't reproduce it myself. I only have crashlogs and they all look the same. But it's affecting quite a lot of my users!!!
@roblav96 do you maybe have some better crashlogs? Or did you manage  to reproduce it?

ccgus:
You don't have the source to the app anymore?

georgbachmann:
Ah... I meant stack trace... give me a sec... I'll post them in a second

roblav96:
Here's the snippet I'm using:
```
FMDatabase *db = [FMDatabase databaseWithPath:@"/tmp/tmp.db"];
if (![db open]) {
    NSLog(@"!db open AWW BOO...")
    [db dealloc];
    return;
}
```
It logs `AWW BOO...` throws
**`fmdb error opening!: 14`**
then crashes.

Let me go find the crash log,

[...]

PR #539, opened 2016-10-04
Update README

robertmryan:
1. Remove hard coded `/tmp` folder. At the very least, use `NSTemporaryDirectory`. You can also use `URLForDirectory` with `NSItemReplacementDirectory`, but I agree [with NSHipster](http://nshipster.com/nstemporarydirectory/) that `NSTemporaryDirectory` is fine.
2. Comment out `[db release]` because for 99% of users, now, this isn't appropriate. But `nil` the pointer either way (in MRC, eliminate dangling pointer reference, in ARC, remove your strong reference).

This is to forestall confusions, such as those raised by roblav96 in issue https://github.com/ccgus/fmdb/issues/538.


ccgus:
Thanks

Issue #542, opened 2016-10-21
How to suppress the log output from fmdb library?


xingheng:
Team, we've integrated this library a long time, could you please provide some solutions to disable the verbose log output from it?

Great thanks for your hard job!

robertmryan:
You can silence some of them by turning off `logsErrors`.

@ccgus - I'm inclined to make all `NSLog` statements honor that setting. Any objection to that? Personally, I'd also be defaulting this to `false` on "release" builds, but IIRC, you weren't crazy about that idea.

xingheng:
Close it as there is an existing PR for it.

PR #543, opened 2016-10-21
Only log errors if `logsErrors` is set

robertmryan:
Pursuant to issue https://github.com/ccgus/fmdb/issues/542, I've updated code to only perform error logging if `logsErrors` is set.

I also updated CHANGES document to reflect some changes over the last few months and I bumped version number so that Carthage and CocoaPods users would pick up this change.

### DIFF
--- a/src/fmdb/FMDatabase.m
+++ b/src/fmdb/FMDatabase.m
@@ -141,7 +141,9 @@
     
     int err = sqlite3_open([self sqlitePath], (sqlite3**)&_db );
     if(err != SQLITE_OK) {
-        NSLog(@"error opening!: %d", err);
+        if (_logsErrors) {
+            NSLog(@"error opening!: %d", err);
+        }
         return NO;
     }
     
@@ -165,7 +167,9 @@
     
     int err = sqlite3_open_v2([self sqlitePath], (sqlite3**)&_db, flags, [vfsName UTF8String]);
     if(err != SQLITE_OK) {
-        NSLog(@"error opening!: %d", err);
+        if (_logsErrors) {
+            NSLog(@"error opening!: %d", err);
+        }
         return NO;
     }
     
@@ -176,7 +180,9 @@
     
     return YES;
 #else
-    NSLog(@"openWithFlags requires SQLite 3.5");
+    if (_logsErrors) {
+        NSLog(@"openWithFlags requires SQLite 3.5");
+    }
     return NO;
 #endif
 }
@@ -203,14 +209,18 @@
                 triedFinalizingOpenStatements = YES;
                 sqlite3_stmt *pStmt;
                 while ((pStmt = sqlite3_next_stmt(_db, nil)) !=0) {
-                    NSLog(@"Closing leaked statement");
+                    if (_logsErrors) {
+                        NSLog(@"Closing leaked statement");
+                    }
                     sqlite3_finalize(pStmt);
                     retry = YES;
                 }
             }
         }
         else if (SQLITE_OK != rc) {
-            NSLog(@"error closing!: %d", rc);
+            if (_logsErrors) {
+                NSLog(@"error closing!: %d", rc);
+            }
         }
     }
     while (retry);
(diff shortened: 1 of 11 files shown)


src/fmdb/FMDatabase.m
```diff
@@ -244,7 +254,7 @@ static int FMDBDatabaseBusyHandler(void *f, int count) {
     if (delta < [self maxBusyRetryTimeInterval]) {
         int requestedSleepInMillseconds = (int) arc4random_uniform(50) + 50;
         int actualSleepInMilliseconds = sqlite3_sleep(requestedSleepInMillseconds);
-        if (actualSleepInMilliseconds != requestedSleepInMillseconds) {
+        if (actualSleepInMilliseconds != requestedSleepInMillseconds && self.logsErrors) {
```
groue: My two cents: without this NSLog, the user won't know something is wrong.

src/fmdb/FMDatabase.m
```diff
@@ -826,7 +843,9 @@ - (FMResultSet *)executeQuery:(NSString *)sql withArgumentsInArray:(NSArray*)arr
                 idx++;
             }
             else {
-                NSLog(@"Could not find index for %@", dictionaryKey);
+                if (_logsErrors) {
```
groue: Without this NSLog, the user won't know she's sending an unknown parameter name.

groue:
Hello @robertmryan. I did not fully review the changes you suggest, but some of them are concerning, when they hide diagnostics that reveal a necessary change in the user code, or a false assumption.

Since FMDB generally does not use error reporting with NSError, and relies on `[db lastErrorMessage]`, you can't really stop logging errors that would not be reported with `[db lastErrorMessage]`.

A proper fix, IMHO, would be to generalize proper NSError reporting. This would also let you create ad-hoc errors whenever the error does not happen inside SQLite. Those NSLogs could then be safely removed.

robertmryan:
@groue - I'm sympathetic to "the developer won't know" issue for some of these cases, but that's clearly only the case if they explicitly turn off `logsErrors`, which defaults to `true`. But it just seems that if a developer explicitly says "don't log any errors", it's strange to continue to log them.

But I agree that a broader review of error handling is definitely warranted. At the very least, it strikes me that a design goal in v3.0 should be that any method that can possibly fail should have mechanism to report the failure in return codes and/or error objects.

groue:
> @groue - I'm sympathetic to "the developer won't know" issue for some of these cases, but that's clearly only the case if they explicitly turn off logsErrors, which defaults to true. But it just seems that if a developer explicitly says "don't log any errors", it's strange to continue to log them.

When "error" means "database error", like a relational constraint failure for example, then you're right, it should not be logged. but when it means "library misuse", then it should not become silent: something has to be done by the developer. "Don't log any error" does not mean "don't tell me when I misuse FMDB".

> But I agree that a broader review of error handling is definitely warranted. At the very least, it strikes me that a design goal in v3.0 should be that any method that can possibly fail should have mechanism to report the failure in return codes and/or error objects.

I find as well that it's too easy to ignore errors with FMDB. But a [sensible error handling](https://github.com/groue/GRDB.swift#error-handling) does not always mean NSError everywhere.

ccgus:
So here's my thoughts- I'm fine with NSLogs not happening where logsErrors is false, but in cases where there is a programmer error, then I think it should log. For instance, in busyRetryTimeout, or where we have "Closing leaked statement".

Cases like columnIndexForName: or resultDict, I can see that looking at the flag before NSLog(ing).

I haven't actually looked over everything here yet though, maybe I should comment again after doing that.

ccgus:
Here's my belated thoughts on this- we'll put this in 3.0. I don't want a coder who is relying on this to all of a sudden have this go away in a minor update.

robertmryan:
I’m going to close this PR as it’s now out of date and will undoubtedly introduce all sorts of merge conflicts. It’s better to just add a “review error logging” to the 3.0 enhancements list, but this particular PR will only add confusion, IMHO. 

@ccgus - Needless to say, feel free to reopen if this helps you keep track of the 3.0 feature list.

PR #549, opened 2016-11-14
Feature-conditional statements for `-finalize` method to opt-out GC methods when using ARC

michalzelinka:
This fixes compilation warnings appearing since Xcode 8 due to deprecated Garbage collection support.

### DIFF
--- a/src/fmdb/FMDatabase.m
+++ b/src/fmdb/FMDatabase.m
@@ -50,10 +50,12 @@
     return self;
 }
 
+#if ! __has_feature(objc_arc)
 - (void)finalize {
     [self close];
     [super finalize];
 }
+#endif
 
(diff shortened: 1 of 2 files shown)


ccgus:
I like this- but do you know if there's anyway to check for the presence of gc, vs. the non-presence of ARC?

michalzelinka:
I haven't found any other convenient way unfortunately. It should still be correct to qualify this against ARC feature as [Apple declares it](https://developer.apple.com/reference/objectivec/nsobject/1418513-finalize) to be called only when using GC. In MRC or ARC envies this method should be ignored (= FMDB only defines it as a local instance method implementation with no implicit call).

ccgus:
Well, it was worth a check. Thanks for the patch!

PR #551, opened 2016-11-15
Patched minor compilation warning for NDEBUG-defined builds

michalzelinka:
Compilation warning: `unused variable `**`currentSyncQueue`**

I've rather used `#ifndef NDEBUG` clause because neither `DEBUG` nor `NDEBUG` need to be defined in the environment and `assert()` is effectively defined the same way in `assert.h`.


ccgus:
Danke!

michalzelinka:
You're welcome! 😉

PR #561, opened 2016-12-30
Llyfmdb

fan19859:
delete  [columnName lowercaseString];


robertmryan:
I assume you discovered why you don't want to do this. By converting to lowercase, we effectively enjoy case insensitivity. I suspect Gus wanted the API to behave like SQLite's standard case insensitive interface, and converting keys to lowercase accomplishes that...

ccgus:
Yep

PR #584, opened 2017-05-26
2.7 - Audit for nullability

robertmryan:
I did the nullability audit in FMDB in the spirit of making FMDB more Swift friendly. See https://github.com/robertmryan/fmdb. This makes it a little more natural for Swift users. I also added file URL constructors because not only has Apple been encouraging us to use URLs for years, now, but they’ve gone so far to actually not implement a bunch of path manipulation methods in Swift, only providing URL manipulation methods.

I also auditing `makeFunctionNamed` routine for as part of this “make it Swifty” effort, because I noticed that you had to drop down the the C API in order to use it, so I created `valueXXX` and `resultXXX` methods to let you make functions more easily for both Objective-C and Swift users. (I also wrapped it in an autorelease pool, so users don’t have to do that anymore. In my experience, you almost always have to use autorelease pools or else a scan through a database could quickly consume an extraordinary amount of memory. Any, see https://github.com/ccgus/fmdb#custom-functions for examples with these new methods.

While I tried to limit non-backward-compatible changes, for Objective-C users, at least, but I did change a three method signatures while I was here (https://github.com/ccgus/fmdb#custom-functions), notably renaming `objectForColumnName` and `UTF8StringForColumnName` to remove the `Name` suffix to be consistent with the other methods. I also changed `makeFunctionNamed` signature a bit, too (the “max arguments” wasn’t really max arguments, but rather the actual number of arguments). In all three cases, I kept the old method there, but marked it as deprecated with a reference to the new method.

There were also a lot of properties that were implemented as simple getter and setter methods, with no formal property definitions, but with manually implemented ivars. In Objective-C this isn't problematic (as the Objective-C "dot" notation is merely syntactic sugar), but it leads to an unintuitive interface for Swift. By shifting these manually implemented getter/setters with manual ivars to properties makes the code a little more intuitive, and leads to more natural looking code in Swift (e.g. rather than `let timeout = db.maxBusyRetryTimeInterval()` and `db.setMaxBusyRetryTimeInterval(value)`, we can just do more natural `let timeout = db.maxBusyRetryTimeoutInterval` and `db.maxBusyRetryTimeInterval = value`. Affected properties include `databasePath`, `maxBusyRetryTimeInterval`, `shouldCacheStatements`, `sqliteHandle`, `hasOpenResultSets`, `lastInsertRowId`, `changes`, `goodConnection`, `columnCount`, `resultDictionary`, `applicationID`, `applicationIDString`, `userVersion`, `countOfCheckedInDatabases`, `countOfCheckedOutDatabases`, and `countOfOpenDatabases`.

While the aforementioned probably sounds scary, it should be entirely transparent to Objective-C users (unless they were using the ivars, egad, that I’ve made private). And for Swift developers, you get a nice simple warning ("remove the '()’").

In addition to the above, I also:

- Updated documentation for the file URL based methods.

- There were a few methods were marked as deprecated with no clear replacement. I’ve used `__deprecated_msg` consistently so that the app developers have a fighting chance to see what the replacement method should be.

- Renamed `isInTransaction`, deprecating its predecessor.

- Fixed a few nullability definitions that had slipped past me on my first go-around, e.g. `stringForColumn` (and all the other similar ones that return pointers), `columnNameForIndex`, etc.

- The `objectForColumn` (and the associated subscript operator) now returns `nil` if an invalid column name/index is passed to it. It used to return `NSNull`. I added unit test for that, too.

- I’ve removed the “Swift variadic” folder from the “Extra” folder. It’s unnecessary, it’s unswifty, and as as separate file you have to include, it was unwieldy to use, anyway.


robertmryan:
Gus replied:

> It looks good to me.
>
> My only nit is that there are places where it accesses properties with self.whatever. At some point in the past I decided to go just use [self whatever] instead of mixing the two different ways in the file. It's not something that should be changed for 2.7, but if I'm in the code and I see one, I'll probably swap it out for [self whatever].
>
> So, I say push this up!
>
> -gus

Per your note, I'll merge this.

ccgus:
Sweet- thanks for all the work Ryan!

Issue #586, opened 2017-05-31
Implicit conversion loses integer precision 'sqlite3_int64' (aka 'long long') to 'long'


Kemenger:
In Xcode 8.3.2 the FMDatabase file method "- (long)valueLong:(void *)value" tips "Implicit conversion loses integer precision 'sqlite3_int64' (aka 'long long') to 'long'"

martinflorek:
How did you solve this? I am getting same error when building v2.7 via Carthage:

`src/fmdb/FMDatabase.m:1495:12: error: implicit conversion loses integer precision: 'sqlite3_int64' (aka 'long long') to 'long' [-Werror,-Wshorten-64-to-32]`

robertmryan:
@Kemenger - I'm trying to reproduce this issue, but you closed this before I got a chance to take a look at it. Does this mean that you found a resolution? I must confess that I'm not seeing this warning. What device were you doing this on? What version of Xcode? Swift or Objective-C app? I'd like to be able to reproduce the warning in order to address it.

@ccgus - Setting aside the fact that I'm having trouble reproducing the warning, there is an inconsistency here. For example, `FMResultSet` method `longForColumnIndex` appears to just silence this warning by casting the result of `sqlite3_column_int64` to `long` before returning it. (That means that if you run it on 32-bit platform, the 64-bit result from `sqlite3_column_int64` is cast to 32-bit result. But it also has a `longLongForColumnIndex` method that returns, obviously, a `long long`.

I'm seeing a couple of approaches:

1. I guess I could do the same for the new `valueLong` as we did for `longForColumnIndex`, e.g. something like:

    <img width="882" alt="screenshot 2017-06-02 11 36 45" src="https://cloud.githubusercontent.com/assets/1530768/26739713/c96f708a-4787-11e7-8ad6-7766a74e4235.png">

    I'm not crazy about perpetuating three permutations, `int`, `long`, and `long long` when SQLite really only has two native types, `int` and `int_64`. This is especially confusing from Swift. For example, look at how Swift renders the existing `FMResultSet` methods `intForColumn`, `longForColumn`, and `longLongIntForColumn`:

    ```swift
    let rs = try! db.executeQuery("select bar from foo", values: nil)
    while rs.next() {
        let a = rs.int(forColumn: "bar")          // returns Swift Int32
        let b = rs.long(forColumn: "bar")         // returns Swift Int
        let c = rs.longLongInt(forColumn: "bar")  // returns Swift Int64
    }
[...]

PR #587, opened 2017-05-31
Fix FMResultSet security weakness

arttrick:
A security weakness was detected as a result of detection by Facebook's infer.
The branch in the code will be taken when the pointer n is non-nil, but the programmer might have actually wanted the branch to be taken when the integer pointed to by n is nonzero. should programmer explicitly compare n to nil or call an accessor to clarify her intention.


ccgus:
The code in FMDB is valid. I don't see how the patch improves things.

robertmryan:
@ccgus - While I agree that this is not needed, I'd actually suggest we go ahead and do it. It's not critical, but it makes the intent more explicit and, more importantly, it eliminates a silly static analyzer warning. It's worth doing just to get a clean bill of health from the static analyzer.

I actually already have this in my pending 2.7.1 release (subject to your feedback, @ccgus, on that `valueLong` question, in issue https://github.com/ccgus/fmdb/issues/586).

ccgus:
The problem with making this change is, that checking for a nil pointer this way is done all over the place in fmdb. I don't have Facebook's infer to try it out, but I'm assuming this isn't the only spot it happens? @arttrick do you see this warning in other places?

robertmryan:
The problem isn't a `nil` pointer. The problem is that it's a `NSNumber *` pointer, and the static analyzer thinks it's ambiguous whether you really meant `if (n != nil) ...` or `if ([n boolValue]) ...`.

screenshot 2017-06-02 16 21 01

By making it `if (n != nil) ...` the ambiguity is removed.

robertmryan:
@ccgus - AFAIK, this is the only `if` statement where we test `NSNumber *` pointer like this. 

Once this ambiguous reference was fixed it, I received a clean bill of health from the static analyzer.

ccgus:
OK, I guess I'm alright with this one off. Is this is a special build of Clang that's using it? I can't get the Xcode analyzer to make the warning

robertmryan:
Yeah, I only stumbled across it when I built a new macOS project in Xcode 8.3.2, and incorporated the FMDB source manually. I suspected some subtle build setting difference, but when I scanned through the build settings of the existing FMDB project and the new project, and nothing jumped out at me. It's funny. 

It's a bit like `valueLong` warning, which I've definitely seen once, but was not able to reproduce. It's frustrating. But in this case the analyzer warning is at least reproducible. If you want me to drop a project out there for you to look at, let me know.

robertmryan:
I've incorporated this fix into 2.7.1, subject to @ccgus approval.

arttrick:
@ccgus This problem is not detected elsewhere.
This problem is not a problem right now. However, there is a problem that the programmer's intentions are not clearly revealed due to unclear comparisons.

ccgus:
And it's committed. Thanks all!

PR #590, opened 2017-06-07
Mark blocks as noescape, if they don't escape the current context

benasher44:
Hello! I read your contributing guidelines, and it said it was best to file an issue to ask questions first. However, I felt this PR was pretty straightforward to implement, so this can serve as my asking if you all would be open to a change like this.

This change marks blocks that do not escape the current context as `noescape`, so that Swift users will not have to mark blocks passed through to FMDB in helper functions as `@escaping`.

Thanks for your consideration!

### DIFF
--- a/src/extra/fts3/FMDatabase+FTS3.h
+++ b/src/extra/fts3/FMDatabase+FTS3.h
@@ -92,7 +92,7 @@ typedef struct FMTokenizerCursor
  using `[FMResultSet columnNameForIndex:]`. The `matchRange` is in UTF-8 byte positions, so it must be 
  modified to use with `NSString` data.
  */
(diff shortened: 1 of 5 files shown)


robertmryan:
I like this! +1

robertmryan:
@ccgus - I'm happy to incorporate this in my forthcoming pull request (where I've updated the headers to expand the documentation) if you want. Or you can merge this and I'll figure out how to incorporate my changes that are inflight with that. I'm happy either way.

benasher44:
Either works for me! 👍

robertmryan:
I've incorporated this into 2.7.2.

benasher44:
Thanks!

ccgus:
Sounds good.

PR #603, opened 2017-08-14
Convert to modern Objective-C syntax

zzd258147:
Xcode contains the "Convert to modern Objective-C syntax" function, but it is difficult to use.

I manually convert the code and test the results.

Please merge my commit, thanks!

### DIFF
--- a/src/fmdb/FMDatabase.m
+++ b/src/fmdb/FMDatabase.m
@@ -114,14 +114,14 @@ NS_ASSUME_NONNULL_END
     dispatch_once(&once, ^{
         NSString *prodVersion = [self FMDBUserVersion];
         
-        if ([[prodVersion componentsSeparatedByString:@"."] count] < 3) {
+        if ([prodVersion componentsSeparatedByString:@"."].count < 3) {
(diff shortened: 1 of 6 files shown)


ccgus:
Thanks for trying this, but I don't see any real reason to change the way things are with it. The syntax it uses is perfectly valid (and as an objc dev for many years, I like using []'s)

zzd258147:
Thanks for your quick reply!

Yes, `[]` syntax is not wrong, but I want FMDB to be simpler and more modern.

Xcode has built-in functions "Convert to modern Objective-C syntax".
(In the Top Menu->Edit->Convert->To Modern Objective-C Syntax)


The changes I made were based on Xcode's automatic conversion, and I only changed two things:
Change `[<Any> count]` to `<Any>.count`;
Change `[NSNumber numberWithInteger:<Any>]` to `@ (<Any>)`.


These two changes are very basic, please reconsider it.🙏🙏🙏

There are some modern features, such as: Nullability, Generic, Designated Initializer. If I have time, I will continue to add these features to FMDB.

PR #606, opened 2017-08-22
Fixed nullability annotations for various FMDatabaseQueue initializers.

robotive:
All the initialisers in the FMDatabaseQueue.h are annotated as returning non nil instances via the NS_ASSUME_NONNULL_BEGIN declaration at the top of the header, yet if you check their implementation (also the documentation comments) they can return nil.

This PR fixes the nullability annotations for the initializers of the FMDatabaseQueue class.

### DIFF
--- a/src/fmdb/FMDatabaseQueue.h
+++ b/src/fmdb/FMDatabaseQueue.h
@@ -86,7 +86,7 @@ NS_ASSUME_NONNULL_BEGIN
  @return The `FMDatabaseQueue` object. `nil` on error.
  */
 
-+ (instancetype)databaseQueueWithPath:(NSString * _Nullable)aPath;
++ (nullable instancetype)databaseQueueWithPath:(NSString * _Nullable)aPath;
 
 /** Create queue using file URL.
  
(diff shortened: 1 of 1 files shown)


benasher44:
Nice! I'd love to see these corrected, since it's otherwise hard to know what can happen nullability-wise from Swift.

robertmryan:
I agree that we should do this. I hate putting in a change that breaks existing Swift code, but the non-optional return values is, as you point out, simply incorrect.

robertmryan:
@ccgus I'm inclined to integrate this pull request, unless you have objections. 

Two additional items to tackle at the same time include (a) updating documentation in README accordingly; and (b) adding tests that demonstrate the failure condition (for the sake of not only coverage, but to confirm that the failable initializers can actually fail.

Do you want me to run with this? 

In terms of tagging this as a release for Carthage/Cocoapods users, I'd be inclined to roll that out at the same time as [other pull request](https://github.com/ccgus/fmdb/pull/609), once he's done reviewing my suggestions, assuming you're in agreement with that one, as well.

ccgus:
@robertmryan Go for it.

benasher44:
🎉 thank you!

PR #609, opened 2017-09-08
Added support for immediate transactions and checkpoint

benasher44:
Prior to FMDB 2.7, I could implement these as categories on `FMDatabase`* classes. Now, most of the internals are no longer exposed in a way that would allow this. This PR seeks to contribute these categories back to FMDB (and allows upgrading code that relied on these categories), which allows using immediate-type transactions as well as running WAL checkpoint (all of these on the FMDatabaseQueue specifically) operations.

Thanks for your consideration!

### DIFF
--- a/Tests/FMDatabaseTests.m
+++ b/Tests/FMDatabaseTests.m
@@ -1464,4 +1464,26 @@
     XCTAssertEqual(error.code, 19, @"error code 19 should have been generated");
 }
 
+- (void)testCheckpoint {
+    FMDatabase *db = [[FMDatabase alloc] init];
+    XCTAssertTrue([db open], @"open failed");
+    NSError *error = nil;
+    int frameCount = 0;
+    int checkpointCount = 0;
+    [db checkpoint:FMDBCheckpointModeTruncate name:NULL logFrameCount:&frameCount checkpointCount:&checkpointCount error:&error];
+    // Verify that we're calling the checkpoint interface, which is a decent scope for this test, without going so far as to verify what checkpoint does
+    XCTAssertEqual(frameCount, -1, @"frameCount should be -1 (means not using WAL mode) to verify that we're using the proper checkpoint interface");
+    XCTAssertEqual(checkpointCount, -1, @"checkpointCount should be -1 (means not using WAL mode) to verify that we're using the proper checkpoint interface");
+}
+
+- (void)testImmediateTransaction {
+    FMDatabase *db = [[FMDatabase alloc] init];
+    XCTAssertTrue([db open], @"open failed");
+    [db beginImmediateTransaction];
+    [db beginImmediateTransaction];
+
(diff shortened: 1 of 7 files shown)


robertmryan:
This looks good. 

A few observations:

1. If we're going to expose `sqlite3_wal_checkpoint_v2` wrapper, I'd suggest a `typedef` for the checkpoint modes rather than `int`. Swift 2/3 users don't have easy access to `sqlite3.h`, so we want to show them their options. It leads to more natural code:

 ```swift
try db.checkpoint(.passive, name: name)
```

2. In the checkpoint method, I might suggest `name` instead of `dbName` (as shown above).

3. If we're providing access to `sqlite3_wal_checkpoint_v2`, we might as well provide access to those last two parameters, too. Let's provide a version with fewer parameters so not everyone is encumbered with that.

4. In the checkpoint method, the name is `_Nullable` (as are the two additional parameters).

I've issued a [pull request](https://github.com/benasher44/fmdb/pull/1) (on your repo) for that so you can see what I was thinking about. Please adjust as you see fit, but just a few thoughts.

I'd also suggest creating some tests for this. At a minimum, we need tests to just run these various methods (so we can keep up our code coverage, make sure we get basic checkpoint behavior, etc.). The question is whether there's some falsifiable test we can do to confirm that this mode offers behavior different than a non-immediate checkpoint.

benasher44:
@robertmryan I added some tests, but I'd say they're a bit unorthodox. They strive to verify that FMDB is somewhat wrapping the interfaces that they advertise to wrap, but they don't go as far as verifying sqlite's actual behavior. I didn't see anything great in sqlite for this, but I think testing could be improved a lot by verifying FMDB's calls to sqlite. Open to thoughts and suggestions here, but I figured this would at least be a good discussion starting point (if not ready to ship as is).

benasher44:
It's been a few weeks. Any thoughts here? Thanks!

robertmryan:
My bad. I’ve just been tied up w real work, but I’ll get to it this week.

benasher44:
No worries. Thanks again!

benasher44:
🎉 thank you!

robertmryan:
@benasher44 - By the way, having just committed this, I thought I'd bring your attention to a question I've asked Gus, et al, that came up while I was doing some additional testing of the above: See https://github.com/ccgus/fmdb/issues/619

benasher44:
Thanks again for merging this! When you have some time, would you mind pushing the updated version to CocoaPods? Thanks!

robertmryan:
Hmm. I created a release for Carthage users, and that looks fine, but when I tried pushing for Cocoapods, it's giving me errors which will take me a little time to diagnose. Stay tuned.

benasher44:
No prob! Let me know if I can be of any help with you he errors.

Issue #619, opened 2017-10-21
Should beginTransaction be exclusive?


robertmryan:
I was working on some test cases for FMDB and noticed that `-[FMDatabase beginTransaction]` and <br />`-[FMDatabaseQueue inTransaction]` both do `EXCLUSIVE` transactions. Given that we want this to be a thin wrapper around SQLite, I don't think this is quite right. The [SQLite documentation](http://sqlite.org/lang_transaction.html) for `BEGIN TRANSACTION` tells us (emphasis added):

> Transactions can be deferred, immediate, or exclusive. The default transaction behavior is _deferred._

I'd suggest:

1. Change the behavior of `beginTransaction` and `inTransaction` to both perform `DEFERRED` transactions, just like `BEGIN TRANSACTION` does; and 

2. Add a new `beginExclusiveTransactions` and `inExclusiveTransaction` (analogous to the existing `beginDeferredTransaction` and `inDeferredTransaction`, but doing exclusive rather than deferred transactions) to offer the `BEGIN EXCLUSIVE TRANSACTION` behavior for those who need it.

I've got a pull request ready to go, but wanted to see if anyone had any objections before I did that. Technically, it changes the behavior of FMDB, but it seems more correct (and it's an edge case, anyway, because most multithreaded contention environments use `FMDatabaseQueue`, which largely obviates the need for the distinctions between the various transaction types).

benasher44:
This makes sense to me! Given the change in behavior, it might be nice to call this out in readme or somewhere else prominent, if the idea is to not do a major version release.

robertmryan:
Good point. If we decide to punt, I might just add the exclusive renditions for now (so people can explicitly adopt the exclusive behavior if they want) and add a warning to the `beginTransaction` and `inTransaction` to note that unlike SQLite’s `BEGIN TRANSACTION`, they do exclusive transactions, subject to change to the more intuitive deferred transactions in the next major release.

benasher44:
Sounds reasonable!

ccgus:
Sorry for the late reply (as usual, I'm a bit slow).

Punt! I think we should add the exclusive renditions, and then a warning to the beginTransaction. That way we don't break any code and the developer will get a nice little warning to know that things are going to change some day.

> "and it's an edge case, anyway, because most multithreaded contention environments use FMDatabaseQueue, which largely obviates the need for the distinctions between the various transaction types)."

[...]

PR #620, opened 2017-10-24
Update transactions

robertmryan:
- Add explicit exclusive transaction code
- Add warning to beginTransaction and inTransaction calls that while it's currently exclusive, we reserve the right to change this in the future
- Add a few tests while I was here


robertmryan:
As discussed in issue https://github.com/ccgus/fmdb/issues/619.

ccgus:
:doubleawesomethumbsup:

PR #629, opened 2017-11-28
Added build target for watchOS

rupey:
Found a watchOS target missing while integrating this great library with Carthage. Hope others find this useful.


ccgus:
This looks reasonable. Thanks for putting it together.

Issue #635, opened 2017-12-18
FMDatabase open error


Eafy:
Hi, I use the 2.7.4 version to print Log: "API call with NULL database connection pointer". I call sequential FMDatabaseQueue -open:, then -close:, and then I can't open database any more. I found that "isOpen" did not set NO in the interface "close", which Open returning to YES again.

Eafy:
[希伯来文.txt](https://github.com/ccgus/fmdb/files/1566896/default.txt)

Eafy:
[Test.txt](https://github.com/ccgus/fmdb/files/1566898/Test.txt)

robertmryan:
With `FMDatabaseQueue`, you don't have to open the database. It's opened for you automatically:

```objc
FMDatabaseQueue *queue = [[FMDatabaseQueue alloc] initWithURL:url];

[queue inDatabase:^(FMDatabase * _Nonnull db) {
    // you don't have to open the database; it's opened for you automatically
    
    [db executeUpdate:@"CREATE TABLE foo (bar TEXT)"];
    [db executeUpdate:@"INSERT INTO foo (bar) VALUES (?)", @"baz"];
    
    // You don't want to close the database this way because `FMDatabaseQueue` maintains its 
    // own database pointer; if you close this, the `FMDatabaseQueue` still has its own dangling
    // pointer to the database you closed, but it has no way of knowing that you closed it
    // unbeknownst to the queue.
    //
    // [db close];
}];

// If you _really_ wanted to close the database, you need to tell the queue to close the database
// it's keeping track of for you. But you generally wouldn't even do this. But if you close the queue's
// database, as shown below, it will now know it's closed and will reopen it when needed.

[queue close];

// Because we told the queue to close the database, it will automatically reopen it again if you
// closed it via `[[queue close]`, above.

[queue inDatabase:^(FMDatabase * _Nonnull db) {
    // this will work fine, having automatically reopened the database for you

    FMResultSet *rs = [db executeQuery:@"SELECT * FROM foo"];
    while ([rs next]) {
        NSLog(@"%@", rs.resultDictionary);
    }
}];
```

Bottom line, certainly never opening the database yourself, because the queue opens it for you. You don't generally close the queue's database, either, but if you needed to, do it like shown above. Frankly, if you have some use case where you think you need to close the queue's database, let us know why, and we might be able to offer better counsel.

Eafy:
Thank you very much. I'm going to modify my code.

PR #636, opened 2017-12-19
Refine close logic

robertmryan:
- If you call `[db close]`, it should reset `isOpen` to false;
- When you try to use `[queue database]`, it should not only check that the `FMDatabase` is non-`nil`, but also whether the database `isOpen` or not;
- Added `testClose` test to `FMDatabaseQueue` logic if the queue's `FMDatabase` was closed.

### DIFF
--- a/Tests/FMDatabaseQueueTests.m
+++ b/Tests/FMDatabaseQueueTests.m
@@ -354,4 +354,19 @@
     
 }
 
+- (void)testClose
+{
+    [self.queue inDatabase:^(FMDatabase *adb) {
+        XCTAssertTrue([adb executeUpdate:@"CREATE TABLE close_test (a INTEGER)"]);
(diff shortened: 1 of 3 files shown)


robertmryan:
Pursuant to issue https://github.com/ccgus/fmdb/issues/635, this prevents a call to `[db close]` in one `-[FMDatabaseQueue inDatabase]` from affecting other `inDatabase` calls elsewhere.

Note, this doesn't change the fact that one shouldn't be closing `db` inside `inDatabase` block, but it was sloppy for me: 

- to not consider closed database in `-[FMDatabaseQueue database]`; as well as 
- to leave `-[FMDatabase isOpen]` true even though the `FMDatabase` was closed. 

The end result is that `isOpen` more accurately reflects the state of the database and `-[FMDatabaseQueue database]` is a tad more robust.

This is a very minor fix, so feel free to merge or disregard as you see fit, @ccgus.

ccgus:
Looks good to me, danke.

PR #637, opened 2017-12-21
Documentation copy/paste fix

sartak:
This was probably copied from FMDatabase.h without updating the record type


ccgus:
Thanks for pointing this out!

PR #645, opened 2018-02-15
Fix for 'This function declaration is not a prototype'.

hankinsoft:
Allows sample code to compile via Xcode 9.2 targeting macOS.


ccgus:
Thanks!

PR #647, opened 2018-02-22
Fixes an issue with double/nsnumber precision

hankinsoft:
Running query:
select round(0.7263,4);

Using
returnValue = [NSNumber numberWithDouble:[self doubleForColumnIndex:columnIdx]];

Results into 0.7262999999999999. This fix gets a string and passes this to a NSDecimalNumber which handles the precision better.

### DIFF
--- a/src/fmdb/FMResultSet.h
+++ b/src/fmdb/FMResultSet.h
@@ -300,6 +300,28 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (NSData * _Nullable)dataForColumnIndex:(int)columnIdx;
 
+/** Result set `NSDecimalNumber` value for column.
+ 
+ @param columnName `NSString` value of the name of the column.
+ 
+ @return `NSDecimalNumber` value of the result set's column.
+ 
+ */
+
+- (NSDecimalNumber * __nullable)decimalForColumn: (NSString *)columnName;
+
+
+/** Result set `NSDecimalNumber` value for column.
+ 
+ @param columnIdx Zero-based index for column.
+ 
+ @return `NSDecimalNumber` value of the result set's column.
+ 
+ */
+
+- (NSDecimalNumber * __nullable)decimalForColumnIndex:(int)columnIdx;
+
+
 /** Result set `(const unsigned char *)` value for column.
 
  @param columnName `NSString` value of the name of the column.
--- a/src/fmdb/FMResultSet.m
+++ b/src/fmdb/FMResultSet.m
@@ -333,6 +333,16 @@
 }
 
 
+- (NSDecimalNumber*)decimalForColumn: (NSString *)columnName {
+    return [self decimalForColumnIndex:[self columnIndexForName:columnName]];
+}
+
+- (NSDecimalNumber*)decimalForColumnIndex:(int)columnIdx {
+    NSString * temp = [self stringForColumnIndex: columnIdx];
+    return [[NSDecimalNumber alloc] initWithString: temp];
+}
+
+
 - (NSData*)dataNoCopyForColumn:(NSString*)columnName {
     return [self dataNoCopyForColumnIndex:[self columnIndexForName:columnName]];
 }
@@ -390,8 +400,7 @@
         returnValue = [NSNumber numberWithLongLong:[self longLongIntForColumnIndex:columnIdx]];
     }
     else if (columnType == SQLITE_FLOAT) {
-        NSString * temp = [self stringForColumnIndex: columnIdx];
-        returnValue = [[NSDecimalNumber alloc] initWithString: temp];
+        returnValue = [NSNumber numberWithDouble:[self doubleForColumnIndex:columnIdx]];
     }
     else if (columnType == SQLITE_BLOB) {
         returnValue = [self dataForColumnIndex:columnIdx];
@@ -428,6 +437,5 @@
 - (id)objectForKeyedSubscript:(NSString *)columnName {
     return [self objectForColumn:columnName];
 }
-
 
 @end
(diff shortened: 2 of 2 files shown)


robertmryan:
I'm sympathetic to the desire to add this capability, but I'd advise against changing the behavior of this existing method. We don't want to break any code reliant upon the existing behavior (e.g. I believe the range of acceptable values captured by a `double` and that of `NSDecimalNumber` are different). Backwards compatibility is of paramount interest, whenever possible.

I might suggest creating a new method for returning a decimal number (e.g. `decimalForColumn` and `decimalForColumnIndex`), but leave the behavior of the existing routines unchanged.

Thoughts?

hankinsoft:
Sure, I'll add a decimalForColumn/Index function. If I were to add a objectForColumnIndex_v2 (leave the original as it is and the _v2 to use the decimal style) would that be ok?

robertmryan:
This is Gus's library, so I'll defer to him, but I'm not an advocate for keeping this revised `objectForColumnIndex` at all. It seems like a pretty "edge case" requirement as so few people use `NSDecimalNumber`. I think it more prudent to offer some special method (`decimalForColumn`/`Index`) for that edge case, and then call it a day.

hankinsoft:
No worries, I'll just go ahead and add the decimalForColumn/Index methods and take care of a special case for objectForColumnIndex outside of the fmdb library. Push will come through shortly.

groue:
May I add my two cents?

SQLite deals with a few core C types, and floats are represented as C doubles.

`[self doubleForColumnIndex:columnIdx]` returns the exact double returned by SQLite: https://github.com/ccgus/fmdb/blob/v2.7/src/fmdb/FMResultSet.m#L278

`[NSNumber numberWithDouble:[self doubleForColumnIndex:columnIdx]]` returns an NSNumber that, if I trust Foundation, stores the exact same double.

As a conclusion, `0.7262999999999999` is a decimal representation of the exact result of `select round(0.7263,4)`.

There is, unless I'm mistaken, absolutely no loss of information here. It just happens that doubles, sometimes, have funny decimal representations. The exact `0.7263` decimal just can not be represented exactly with a double. Neither do `0.1` or `0.2`. Try to display the result of `SELECT 0.1 + 0.2`, for example -> `0.3000000000000000512`, yikes!

That's for the general answer. Now, the `-[decimalForColumn:]` and `-[decimalForColumnIndex:]` methods can come in handy. But they won't quite fix anything, since there is nothing to be fixed.

robertmryan:
I agree it's not "fixing" anything. It's just a new feature. If you had a decimal number saved as a string in your database, you could now properly retrieve it more easily with `decimalWithColumn`. 

In the spirit of "fixes", though, what about saving values to the database? I'm actually thinking that it might be good to modify `bindObject:toColumn:inStatement:` to test for `NSDecimalNumber` and, if so, write it as a string to the database, so that FMDB doesn't take a a nice accurate decimal number and then introduce the problems of saving it as a `double`. Thoughts?

robertmryan:
@groue - By the way, I didn't mention it, but I think your assessment of his `select round(0.7263,4)` scenario is spot on. That's a red herring. It's repeating a common problem, conflating "rounding" with the challenges of representing decimal numbers accurately in binary floating point types.

I was trying to answer the more meaningful question: How does one store and retrieve a `NSDecimalNumber` into a database without involving `double` at any point in the process? I think `decimalForColumn` is a good start, but only tackles half of that equation.

ccgus:
@robertmryan has already written out everything that I was thinking, so I won't repeat anything there.

Adding `decimalForColumn` + watching for `NSDecimalNumber` when binding would be a reasonable solution to all of this, and I can't think of any backwards compatibility problems.

groue:
> I was trying to answer the more meaningful question: How does one store and retrieve a NSDecimalNumber into a database without involving double at any point in the process? I think decimalForColumn is a good start, but only tackles half of that equation.

Spot on, Rob

> If you had a decimal number saved as a string in your database, you could now properly retrieve it more easily with decimalWithColumn.

Thanks for this excellent idea!

> In the spirit of "fixes", though, what about saving values to the database? I'm actually thinking that it might be good to modify bindObject:toColumn:inStatement: to test for NSDecimalNumber and, if so, write it as a string to the database, so that FMDB doesn't take a a nice accurate decimal number and then introduce the problems of saving it as a double. Thoughts?

Storing decimals as strings is a good technique indeed, with the caveat that this breaks sorting (unless you zero-pad), and aggregate operations like sums (unless you provide a custom aggregate function).

Developers who have to deal with currency amounts often store decimals as integers (don't store dollars, store cents instead), in order to preserve exact computations, ordering, and aggregates.

Storing NSDecimalNumber as SQLite numbers is bound to lose information, for decimals that can't be exactly represented as a double or an int64. Yet we can improve the situation for big integers that fit in int64, but can't be represented as doubles, by storing NSDecimalNumber as integers when possible, and doubles otherwise ([sample code](https://github.com/groue/GRDB.swift/blob/v2.8.0/GRDB/Core/Support/Foundation/NSNumber.swift#L11-L17)). You end up with a database column that contains a mix of int64 and double. You can still sort, aggregate, and you have "more" exact values. You need a decoding method that handles mixed data types in order to preserve as much information as possible.

To sum up, my humble opinion is that one has to choose if a library is an SQLite wrapper, or a high-level data management system that happens to use SQLite (and hides all its warts). I've always seen FMDB as an SQLite wrapper. Learning to properly use FMDB involves, sometimes, learning about SQLite. Eventually one has to learn about decimals, doubles, integers, and strategies to store decimals.

groue:
> I can't think of any backwards compatibility problems.

When reading a database value that contains `3.0e+5` (the string), sqlite3_column_double returns 30000 (the double). That's because SQLite performs string to number conversion when asked to. FMDB lets this conversion leak in `doubleForColumnIndex:`. So I guess we want to have `decimalForColumn:` return 30000 as well.

And how lucky are we? `NSDecimalNumber(string: "3.0e+5")` is consistent with SQLite, and returns 30000. "2018-02-24" is OK, too. Looks like there aren't much corner cases (but I haven't searched more) :-)

PR #649, opened 2018-02-26
Always use inDatabase: in FMDatabaseQueue. Respect crashOnErrors for open

benasher44:
In our code, we use a similar `dispatch_get_specific` technique to guard against deadlock. When I realized this, I was hoping I could just use what's built into FMDB instead. However, It appears not all calls go through `inDatabase:`, which means those other calls could deadlock un-knowlingly (or at least without making much noise).

In doing that, I also realized that you could conceivably call `close` on an `FMDatabaseQueue` and then call `inDatabase:` and have the `[self database]` fail to re-open the database and call the block with a `nil` db, which would break the nullability guarantee of `nonnull` as described in `FMDatabaseQueue.h`. This isn't a big deal in Obj-C, but in Swift it'll cause crash. I fixed it to bail out early and not call the block if the re-open fails. I also added `crashOnErrors` support to the `open` methods in `FMDatabase` to make this scenario hard to miss in debug.

After writing this, it occurs to me that there could have been some reason for not `re-using` `inDatabase:` everywhere. If that's the case, I'm open to refactoring this in a different way to accomplish the same goal. Thanks!

### DIFF
--- a/src/fmdb/FMDatabase.m
+++ b/src/fmdb/FMDatabase.m
@@ -175,6 +175,9 @@ NS_ASSUME_NONNULL_END
 
     int err = sqlite3_open([self sqlitePath], (sqlite3**)&_db );
     if(err != SQLITE_OK) {
+        if (_crashOnErrors) {
+            NSAssert(false, @"The FMDatabase could not be opened (err: %d) at path: %s", err, [self sqlitePath]);
+        }
         NSLog(@"error opening!: %d", err);
         return NO;
     }
@@ -209,6 +212,9 @@ NS_ASSUME_NONNULL_END
     
     int err = sqlite3_open_v2([self sqlitePath], (sqlite3**)&_db, flags, [vfsName UTF8String]);
     if(err != SQLITE_OK) {
+        if (_crashOnErrors) {
+            NSAssert(false, @"The FMDatabase could not be opened (err: %d) at path: %s", err, [self sqlitePath]);
+        }
         NSLog(@"error opening!: %d", err);
         return NO;
     }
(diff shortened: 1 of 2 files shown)


cacaosteve:
Would this help a crash I am getting using the database queue from my main thread? It only happens sometimes. It's a background thread that is crashing. I know I could use a dispatch queue from the main thread, but also that blocking should work fine.

libobjc.A.dylib
objc_release
libobjc.A.dylib
object_cxxDestructFromClass(objc_object*, objc_class*)
libobjc.A.dylib
objc_destructInstance
libobjc.A.dylib
_objc_rootDealloc
Foundation
-[NSOperation dealloc]
libsystem_blocks.dylib
bool HelperBase<GenericInline>::disposeCapture<(HelperBase<GenericInline>::BlockCaptureKind)3>(unsigned int, unsigned char*)
libsystem_blocks.dylib
HelperBase<GenericInline>::destroyBlock(Block_layout*, bool, unsigned char*)
libsystem_blocks.dylib
_call_dispose_helpers_excp
libsystem_blocks.dylib
_Block_release
libdispatch.dylib
__destroy_helper_block_8_32c35_ZTS29dispatch_block_private_data_s
libsystem_blocks.dylib
_call_dispose_helpers_excp
libsystem_blocks.dylib
_Block_release
libdispatch.dylib
_dispatch_client_callout
libdispatch.dylib
_dispatch_continuation_pop
libdispatch.dylib
_dispatch_async_redirect_invoke
libdispatch.dylib
_dispatch_root_queue_drain
libdispatch.dylib
_dispatch_worker_thread2
libsystem_pthread.dylib
_pthread_wqthread
libsystem_pthread.dylib
start_wqthread

lzcuriosity:
您的邮件已经收到，我会尽快查阅，谢谢！Best Wishes!

PR #651, opened 2018-03-01
Update nullability annotations for `-[FMDatabase getSchema]` and `-[FMDatabase getTableSchema]`

amorde:
The schema functions return the result of calling `-[FMDatabase executeQuery:withArgumentsInArray:orDictionary:orVAList:]`

Since `-[FMDatabase executeQuery:withArgumentsInArray:orDictionary:orVAList:]` can return nil on error, the type annotation of these functions should match that.

The function doc also mentions that this function can return nil:

>  @return `FMResultSet` of schema; `nil` on error.

This is important for Swift, as attempting to access a nil result which is currently typed as `FMResultSet` (instead of `FMResultSet?) will crash.

### DIFF
--- a/src/fmdb/FMDatabaseAdditions.h
+++ b/src/fmdb/FMDatabaseAdditions.h
@@ -142,7 +142,7 @@ NS_ASSUME_NONNULL_BEGIN
  @see [SQLite File Format](http://www.sqlite.org/fileformat.html)
  */
 
-- (FMResultSet *)getSchema;
+- (FMResultSet * _Nullable)getSchema;
 
 /** The schema of the database.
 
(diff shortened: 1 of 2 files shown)


src/fmdb/FMDatabaseAdditions.h
```diff
@@ -166,7 +166,7 @@ NS_ASSUME_NONNULL_BEGIN
  @see [table_info](http://www.sqlite.org/pragma.html#pragma_table_info)
  */
 
-- (FMResultSet*)getTableSchema:(NSString*)tableName;
+- (FMResultSet * _Nullable)getTableSchema:(NSString*)tableName;
```
amorde: For the spacing / formatting of the return type, I referenced other functions that have nullability annotations such as this one:

https://github.com/ccgus/fmdb/blob/240667c4d47b438458b917424150bc07e8c644b1/src/fmdb/FMDatabase.h#L608

ccgus:
Looks good, thanks!

amorde:
Thanks for the quick merge!

PR #668, opened 2018-05-07
join issue: in  FMResultSet

chdo002:
added a fixed property:   fat_resDictionary

take care of the condition :  join two tables, and they share some column names, in some cases, we may get Nsnull for these columns.

e.g. :
table A    colum1: 123,  colum2:234 , colum3:345

table B    colum2:234, colum3:null

select * from tableA left join table B on tableA.colum2 = tableB.colum2


ccgus:
This would probably be better in a category of your own creation for now. What does the signature "fat_resDictionary" mean?

chdo002:
ok

Issue #618, opened 2017-10-16
FMDB/SQLCipher issue on Xcode 9


zhoufusong:
s.dependency 'FMDB/SQLCipher', '~> 2.7.2' in podspec of other pod library. it cannot build successful. also the reason is that sqlite3_key and  sqlite3_rekey is missed

cyl123:
我也遇到了这个问题，解决了么？

suweiz19890921:
I have this problem too.

R4N:
You could try this post_install hook in your Podfile to add SQLCipher to the `HEADER_SEARCH_PATHS` which will ensure that FMDB finds the correct SQLCipher copy of sqlite3.h

```
post_install do |installer|
        installer.pods_project.targets.each do |target|
            if target.name == "FMDB"
               target.build_configurations.each do |config|
                   header_search = {"HEADER_SEARCH_PATHS" => "SQLCipher"}
                   config.build_settings.merge!(header_search)
               end
            end
        end
    end
```

I've issued a PR which should hopefully accomplish the same thing within the SQLCipher subspec of the podspec.

R4N:
https://github.com/ccgus/fmdb/commit/18152d3ea819d5a65a682ad9f3b22dbbb3c2423e was merged in which I'm hopeful will resolve this issue.

@zhoufusong @cyl123 @suweiz19890921

Does it work for you guys with the latest?

Issue #612, opened 2017-09-20
run on Xcode9 


furueili:
![2017-09-20 2 34 52](https://user-images.githubusercontent.com/1407724/30629936-ee7a7d2c-9e10-11e7-8f0d-b5b71eb94d38.png)

just update Xcode9 and found this issue please help thx

mezhevikin:
H @furueili !
Fmdb does not support xcode9 + ios11 yet.
I'm waiting fix too.

billymeltdown:
Are you using Cocoapods to include FMDB in your project, and are you also attempting to use SQLCipher? If so, you may find this helpful:

https://discuss.zetetic.net/t/ios-11-xcode-issue-implicit-declaration-of-function-sqlite3-key-is-invalid-in-c99/2198/27

The error suggests `sqlite3.h` is not in your header search paths (or, that the `sqlite3.h` that is being used is not the SQLCipher version including the symbols `sqlite3_key`, `sqlite3_rekey`, etc)

mezhevikin:
hi @billymeltdown . Thank you for answer. Yes, i use SQLCipher.
I tried it hack, and application was compiled.
But i'm getting crash:

`libc++abi.dylib: terminating with uncaught exception of type NSException`

My podfile:

```
source 'https://github.com/CocoaPods/Specs.git'
platform :ios, '8.0'
use_frameworks!

abstract_target 'defaults' do
    pod 'SQLCipher', :inhibit_warnings => true
    pod 'FMDB/SQLCipher', :git => 'https://github.com/ccgus/fmdb'
    target 'DB'
    target 'Example'
end

post_install do | installer |
    print "SQLCipher: link Pods/Headers/sqlite3.h"
    system "mkdir -p Pods/Headers/Private && ln -s ../../SQLCipher/sqlite3.h Pods/Headers/Private"
end

```

mezhevikin:
@billymeltdown thk.
Hack works fine! Problem was with my code.

R4N:
Unfortunately, this post_install hook no longer seems to do the trick with the latest version of cocoa pods and Xcode.  This post_install hook to add SQLCipher to the HEADER_SEARCH_PATHS of FMDB should do the trick now:

```
post_install do |installer|
        installer.pods_project.targets.each do |target|
            if target.name == "FMDB"
               target.build_configurations.each do |config|
                   header_search = {"HEADER_SEARCH_PATHS" => "SQLCipher"}
[...]

Issue #591, opened 2017-06-12
Can't build on Xcode 9.0


Boobby69:
Hi ! 

I can't build FMDB with this pod file line :
   pod 'FMDB/SQLCipher' with FMDB v. 2.7.2 & SQLCipher v. 3.4.1

The error I get in class FMDatabase.m is : 
> Implicit declaration of function 'sqlite3_rekey' is invalid in C99 

I waited for new version (I was in v. 2.6) and still have the same problem.

FYI, it works with Xcode 8.3.3.

Thanks

andrewcl:
This issue can be resolved by modifying the import statement for sqlite3.h in the FMDatabase.m at line 8 from the following 
`#import <SQLCipher/sqlite3.h>`
to this
`#import <sqlite3.h>`

cyvoit:
@andrewcl change `#import <sqlite3.h>` to`#import <SQLCipher/sqlite3.h>` works for me. Thanks.

MohammadHejazi:
If you are using a static library for SQLCipher, there will be no `<SQLCipher/sqlite3.h>`. In this case you can declare the missing functions by adding the following to `FMDatabase.m`:
```
#if defined(SQLITE_HAS_CODEC)
SQLITE_API int sqlite3_key(sqlite3 *db, const void *pKey, int nKey);
SQLITE_API int sqlite3_rekey(sqlite3 *db, const void *pKey, int nKey);
#endif
```
Or you can use this fork temporarily until the issue is fixed:
```
pod 'FMDB', :git => 'git@github.com:BayanLabs/fmdb.git', :commit => 'eea8a334'
```

ricellis:
@ccgus is there an outlook on #594 merging and making it into a release to resolve this issue? Or an outlook or plan for some alternative fix?

billymeltdown:
These problems sound a lot like what is described in this advisory on using SQLCipher in Xcode 9: https://discuss.zetetic.net/t/important-advisory-sqlcipher-with-xcode-9-and-new-sdks/

It may be the case that by following the guidance there no adjustment to FMDB is necessary (e.g. `#import <sqlite3.h>` may work fine), depending on your individual circumstances and how you link against SQLCipher. For instance, if you are using CocoaPods and `use_frameworks!` you'll want to use a post install hook to make the generated `sqlite3.h` visible (see link). Hope that helps!

ccgus:
[...]

PR #672, opened 2018-05-17
Adds SQLCipher to the HEADER_SEARCH_PATHS of FMDB within the SQLCiphe…

R4N:
…r subspec.

When installing FMDB/SQLCipher via cocoa pods and using Swift, the proper SQLCipher copy of sqlite3.h isn't located so the sqlite3_key and sqlite3_rekey functions aren't found.  

This should hopefully resolve these issues:
https://github.com/ccgus/fmdb/issues/591
https://github.com/ccgus/fmdb/issues/618
https://github.com/ccgus/fmdb/issues/612

and accomplish the same thing as this PR (without changing the import statements/breaking direct integrations):
https://github.com/ccgus/fmdb/pull/594

A current workaround is to use a post_install hook in the Podfile:

```
post_install do |installer|
        installer.pods_project.targets.each do |target|
            if target.name == "FMDB"
               target.build_configurations.each do |config|
                   header_search = {"HEADER_SEARCH_PATHS" => "SQLCipher"}
                   config.build_settings.merge!(header_search)
               end
            end
        end
    end
```

### DIFF
--- a/FMDB.podspec
+++ b/FMDB.podspec
@@ -43,8 +43,7 @@ Pod::Spec.new do |s|
     ss.dependency 'SQLCipher'
     ss.source_files = 'src/fmdb/FM*.{h,m}'
     ss.exclude_files = 'src/fmdb.m'
-    ss.xcconfig = { 'OTHER_CFLAGS' => '$(inherited) -DSQLITE_HAS_CODEC -DHAVE_USLEEP=1' }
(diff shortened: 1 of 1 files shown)


ccgus:
Crossing my fingers and hoping for the best here. I don't use Cocoa Pods, so…

PR #673, opened 2018-05-21
-[FMDatabase executeStatements:withResultBlock:] method crash

rob2468:
`-[FMDatabase executeStatements:withResultBlock:]` method has potential crash bugs, for setting nil object as NSMutableDictionary value

### DIFF
--- a/Tests/FMDatabaseTests.m
+++ b/Tests/FMDatabaseTests.m
@@ -1151,12 +1151,24 @@
         XCTAssertEqual(count, 1, @"expected one record for dictionary %@", dictionary);
         return 0;
     }];
+    
+    XCTAssertTrue(success, @"bulk select");
 
+    // select blob type records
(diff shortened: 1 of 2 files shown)


ccgus:
So the problem is that a blob is trying to be loaded into an NSString (just going over the code and failure points). So the check for values[i] ? … : …  passes, we get a nil value back anyway. Neat.

@robertmryan Any opinions on the patch? Looks reasonable.

rob2468:
It has been more than ten days since this PR submitted. Any questions?

ccgus:
It's merged. I wanted to take a closer look at it, since I wasn't originally sure why it was failing. But it makes sense now, since the bytes that were being loaded into an NSString weren't compatible with it, so it returned nil.

rob2468:
Yes, it is. The PR includes a test case that insert a binary file into table, then access with the objective method. It does crash as expected.

PR #677, opened 2018-06-05
Added attributes to method implementations

philippec:
In Xcode 10, if the `__attribute__((noescape))` is present in the header file,
it must also be present in the implementation.

I have verified that all targets build and all unit tests pass.

### DIFF
--- a/src/fmdb/FMDatabase.m
+++ b/src/fmdb/FMDatabase.m
@@ -1272,7 +1272,7 @@ int FMDBExecuteBulkSQLCallback(void *theBlockAsVoid, int columns, char **values,
     return [self executeStatements:sql withResultBlock:nil];
 }
 
-- (BOOL)executeStatements:(NSString *)sql withResultBlock:(FMDBExecuteStatementsCallbackBlock)block {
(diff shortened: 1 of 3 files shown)


philippec:
Not that this does not update to Xcode 10, there are still warnings when building. But it does build and run. Without this change you can't even build.

yeahphil:
+1 -- I started to do this, too before I saw your work. Thanks for putting together the pull request!

mattneub:
Thanks for doing this. I hope this gets pulled into the base as soon as possible.

ccgus:
Thanks @philippec !

a2:
Are you planning to do a new release soon with these changes?

ccgus:
I wasn't planning on doing one ASAP. Is it needed for CocoaPods integration or something?

a2:
@ccgus We're currently pointing our Carthage build at 61828b418ae9d52fb861ad91fc1226f485379832. This is fine since newer Xcode 10 beta builds may break FMDB in further unexpected ways. So I think it's fine not to tag a release until Xcode 10 stabilizes :)

PR #678, opened 2018-06-07
Annotate method definitions with __attribute__((noescape)), …

ancientbuho:
… not just declarations

This fixes the compiler error:
Parameter of overriding method should be annotated with __attribute__((noescape))


mattneub:
Identical to https://github.com/ccgus/fmdb/pull/677

ancientbuho:
@mattneub ok - i'll close it

PR #690, opened 2018-08-01
Replace __attribute__((noescape)) with a macro to support compiling it with not supported compilers

maicki:
Hey! For linting / static analysis we are running a custom version of clang from llvm.org. Before clang 8, `__attribute__((noescape))` is not available and therefore if we are upgrading to the latest version of fmdb (or latest master) it will error out while compiling.

This pr is adding a macro to only add the attribute if the compiler supports it. We already did the same on `https://github.com/pinterest/PINRemoteImage` with `PIN_NOESCAPE`

I chose `fmdb_noescape` for the name, but I'm open to change it to whatever :)

### DIFF
--- a/src/fmdb/FMDatabase.h
+++ b/src/fmdb/FMDatabase.h
@@ -521,7 +521,7 @@ typedef NS_ENUM(int, FMDBCheckpointMode) {
 
  */
 
-- (BOOL)executeStatements:(NSString *)sql withResultBlock:(__attribute__((noescape)) FMDBExecuteStatementsCallbackBlock _Nullable)block;
+- (BOOL)executeStatements:(NSString *)sql withResultBlock:(fmdb_noescape FMDBExecuteStatementsCallbackBlock _Nullable)block;
 
 /** Last insert rowid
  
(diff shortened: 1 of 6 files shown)


ccgus:
Why not update your custom version of clang? That seems like a much more reasonable solution, rather than modifying FMDB for a single case. Or, run FMDB through a sed to remove `__attribute__((noescape))` before running through your custom clang.

maicki:
@ccgus Sounds good to me - Thanks!

CodeStage:
I have the same error with Xcode 10.0 beta 6 and I don't have a custom version of clang.

@ccgus Maybe you want to reconsider, because the current FMDB version can't be installed with Carthage because of that...

CodeStage:
Here is the actual error message:

'parameter of overriding method should be annotated with __attribute__((noescape))' [-Werror,-Wmissing-noescape]

CodeStage:
Sorry I got that wrong, the error with with Xcode is actually because some annotations are missing, not because the attribute is not available 🙈

PR #696, opened 2018-08-14
Mark the setDateFormat method's parameter as _Nullable

jfahrenkrug:
Passing in nil resets the FMDB to use a default unix date format. At the moment calling
this method with nil raises a warning in Xcode because it is marked as nonnull
because of the NS_ASSUME_NONNULL_BEGIN marcro.

This commit marks it correctly as _Nullable.


ccgus:
Thanks for the patch.

jfahrenkrug:
@ccgus Thanks so much for an outstanding piece of open source software. Oh, and thanks for the quick merge :]

PR #697, opened 2018-08-14
Use NSLocalizedStringFromTable

jhammer:
For apps that include FMDB source in their Xcode projects (instead of linking against the framework), it may be nice to place all of FMDB's localized strings in their own strings table instead of having them pollute the global Localizable.strings table.


ccgus:
Looks reasonable, thanks.

PR #704, opened 2018-09-04
Add Bazel build rules.

russellhancox:
This allows other projects using Bazel (http://bazel.build) to easily import
FMDB by adding it to their WORKSPACE. 

I've included, from the CocoaPods spec, rules for core FMDB and the 
FTS extensions but not the rules that had extra dependencies. There are
also rules for running the unit tests from Bazel, in case that's useful.

I also made a capitalization change in FMDatabase+FTS3.h to silence a compiler warning.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ build
 fmdb.xcodeproj/*.mode1v3
 *.xcodeproj/project.xcworkspace/xcuserdata/
 fmdb.xcodeproj/project.xcworkspace
+bazel-*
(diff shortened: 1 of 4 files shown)


ccgus:
I've got the change for the compiler warning up, thanks for that.

I've not heard of Bazel before, and I'm not inclined to add support for things that we don't use normally, sorry. FMDB already has CocoaPods support, and that never gets updated because I don't use it and know next to nothing about it :/

russellhancox:
I thought that might be the case but figured I'd give it a shot. Thanks anyway!

PR #707, opened 2018-09-18
Feature - Xcode 10 Support

cnoon:
This PR updates the Xcode project to support the Xcode 10 GM without any warnings. All changes are backwards compatible with previous Xcode versions back to Xcode 8.0.

The specific changes are as follows:

- Updated the project accepting all the suggested changes from Xcode.
- Bumped the compatibility version from 3.2 to 8.0
- Re-added the test files to the project to fix up a localization warnings around the `InfoPlist.strings` file generated for the test target

These changes will make it easier for clients to use FMDB as a submodule with Xcode 10. Thank you for your project and continuing to maintain it over time! 🍻


ccgus:
Thanks!

PR #714, opened 2018-11-05
Improvements for executing query and update.

iwill:
Improvements for executing query and update:
1. sql supports parameter prefix ":" or "$" or "@" which supports by sqlite3;
2. bind null if `dictionaryArgs` does not contain a key from `pStmt`;
3. not to bind if `pStmt` does not contain a key of `dictionaryArgs`;

The 2 and 3 are useful when `dictionaryArgs` is a JSON object, which serialized from an Objective-C object. Some keys need not be stored into database, and some other keys are ignored since their values are nil.

With this commit, we donot need create another dictionary which must has the apropos key-values any more.


ccgus:
At first glance, this looks reasonable and tests pass. Let me look at it a bit more though before I make a commit.

PR #725, opened 2019-01-18
issuefix, dataBuffer type fix

Lision:


ccgus:
Do you have an example where this is a problem?

Lision:
@ccgus Emmmmm... Nope, I just fund the `sqlite3_column_blob` return type is `void *`, so I think there is no need to change it to `char *`...

Issue #724, opened 2019-01-04
 [logging] BUG IN CLIENT OF sqlite3.dylib: illegal multi-threaded access to database connection


edsgershao:
```
(lldb) bt all
* thread #1, queue = 'fmdb.<FMDatabaseQueue: 0x6000003fd5f0>', stop reason = EXC_BAD_INSTRUCTION (code=EXC_I386_INVOP, subcode=0x0)
    frame #0: 0x000000010f4accc5 libsqlite3.dylib`sqlite3MutexMisuseAssert + 117
    frame #1: 0x000000010f4acb16 libsqlite3.dylib`checkMutexEnter + 38
    frame #2: 0x000000010f4900a3 libsqlite3.dylib`sqlite3LockAndPrepare + 211
    frame #3: 0x000000010f48196b libsqlite3.dylib`sqlite3_prepare_v2 + 43
  * frame #4: 0x000000010a3f0001 MicroVideo`-[FMDatabase executeQuery:withArgumentsInArray:orDictionary:orVAList:](self=0x0000600002994c60, _cmd="executeQuery:withArgumentsInArray:orDictionary:orVAList:", sql=@"SELECT * FROM send_packet", arrayArgs=0x0000000000000000, dictionaryArgs=0x0000000000000000, args=0x00007ffee5d6e900) at FMDatabase.m:833
    frame #5: 0x000000010a3f0dba MicroVideo`-[FMDatabase executeQuery:](self=0x0000600002994c60, _cmd="executeQuery:", sql=@"SELECT * FROM send_packet") at FMDatabase.m:953
    frame #6: 0x000000010aa6c12a MicroVideo`__29-[MFSendPacketDao getPackets]_block_invoke(.block_descriptor=0x00007ffee5d6ec90, db=0x0000600002994c60) at MFSendPacketDao.m:62
    frame #7: 0x000000010a3faa9d MicroVideo`__30-[FMDatabaseQueue inDatabase:]_block_invoke(.block_descriptor=0x00007ffee5d6ebf0) at FMDatabaseQueue.m:194
    frame #8: 0x0000000118109602 libdispatch.dylib`_dispatch_client_callout + 8
    frame #9: 0x0000000118117653 libdispatch.dylib`_dispatch_lane_barrier_sync_invoke_and_complete + 132
[...]

Issue #711, opened 2018-10-16
crash: Unlock of an os_unfair_lock not owned by current thread in iOS12


dongfanggugu:
the iOS version is 12.0.0
the FMDatabaseQueue is the single instance in my app
i call as 
 dispatch_async(concurrencyQueue, ^{
     FMDatabaseQueue *queue = [[SMDataBaseManager sharedInstance] getDBQueue];
     [queue inTransaction:^(FMDatabase *db, BOOL *rollback) {
            //do something
     }];
});

sometimes, the app crashed, report "Unlock of an os_unfair_lock not owned by current thread"
call stack:

libsystem_platform.dylib | _os_unfair_lock_unowned_abort + 36
-- | --
1 | libsystem_platform.dylib | _os_unfair_lock_unlock_slow + 144
2 | libsqlite3.dylib | sqlite3_snprintf + 3296
3 | libsqlite3.dylib | sqlite3_snprintf + 2592
4 | libsqlite3.dylib | sqlite3_free_table + 48616
5 | libsqlite3.dylib | sqlite3_bind_int64 + 1356
6 | libsqlite3.dylib | sqlite3_free_table + 60928
7 | libsqlite3.dylib | sqlite3_step + 28796
8 | libsqlite3.dylib | sqlite3_step + 432
9 | FMDB | FMDatabase.m line 1073-[FMDatabase executeUpdate:error:withArgumentsInArray:orDictionary:orVAList:]
10 | FMDB | FMDatabase.m line 1158-[FMDatabase executeUpdate:]

the FMDB version is '2.6.2'

Tomsocket:
the iOS version is 12.0.0
the FMDatabaseQueue is the single instance in my app


[logging] BUG IN CLIENT OF sqlite3.dylib: illegal multi-threaded access to database connection

ccgus:
@Tomsocket Show a backtrace of all your threads please.

wnk77521:
![image](https://user-images.githubusercontent.com/12749341/47403919-e8b72280-d77d-11e8-9915-fee192a46429.png)
@ccgus hi  i have  the same  crash  ,   thinks

Mr-yuwei:
me too

ccgus:
Show a back trace of _all_ your threads. Copy and paste them in here (no screenshots please)

dongfanggugu:
//the crash trace
# Platform: ios
# Application: 
# Version: 7.2.1 (801809301648)
[...]

PR #728, opened 2019-01-28
fix issue #724 & #711

crmo:
I write a demo to recurrent issue #724 & #711：

```objc
    NSString *dbPath = [docPath stringByAppendingPathComponent:@"test.sqlite"];
    _queue = [FMDatabaseQueue databaseQueueWithPath:dbPath];
    for (int i = 0; i < 10; i++) {
        dispatch_async(dispatch_get_global_queue(0, 0), ^{
            [_queue inDatabase:^(FMDatabase * _Nonnull db) {
                FMResultSet *result = [db executeQuery:@"select * from test1 where a = '1'"];
                if ([result next]) {
                }
                // do not close FMResultSet
//                [result close];
            }];
        });
    }
```

I think the crash reason was if user traversed FMResultSet used `if ([result close])` not `while ([result close])`. 
FMResultSet `dealloc` in `AutoreleasePoolPage::pop()`, then `[FMResultSet close]` will be called, if user  reading and writing to the database currently, the database was accessed by multi-threaded.
I suggest to modify this like this, if the user did not close the FMResultSet after the block ends, we close it in current dispatch_queue_t.

### DIFF
--- a/src/fmdb/FMDatabaseQueue.m
+++ b/src/fmdb/FMDatabaseQueue.m
@@ -195,6 +195,7 @@ static const void * const kDispatchQueueSpecificKey = &kDispatchQueueSpecificKey
         
         if ([db hasOpenResultSets]) {
             NSLog(@"Warning: there is at least one open result set around after performing [FMDatabaseQueue inDatabase:]");
+            [db closeOpenResultSets];
             
 #if defined(DEBUG) && DEBUG
             NSSet *openSetCopy = FMDBReturnAutoreleased([[db valueForKey:@"_openResultSets"] copy]);


crmo:
The back trace

crmo:
@ccgus Thanks for your attention, I added a unit test for this problem. 

When I readed the FMDB`s document, in chapter `Executing Queries` , I saw a description of this issue.

> You must always invoke -[FMResultSet next] before attempting to access the values returned in a query, even if you're only expecting one:

```objc
FMResultSet *s = [db executeQuery:@"SELECT COUNT(*) FROM myTable"];
if ([s next]) {
    int totalCount = [s intForColumnIndex:0];
}
```

> Typically, there's no need to -close an FMResultSet yourself, since that happens when either the result set is deallocated, or the parent database is closed.

So I used `if ([s next])` because I just need one value, but application was crash. Is it misleading here?

Usually we will use `while ([s next])` to iterate, `[FMResultSet close]` will called in the end, this is same as call `[FMDatabase closeOpenResultSets]` in `FMDatabaseQueue` when block completed. We need to ensure that the `FMDatabaseQueue` is fully thread safe, or explain the risks in the documentation.

zhouyantongiosDev:
@crmo Do you have contact information, such as QQ?

CodeLife2012:
Cached stat access by multi thread. Root cause maybe gcd autorelease pool pop is not invoked when finish block.

CodeLife2012:
Maybe need to assert on the FMResultSet should dealloc on the private queue.

Issue #736, opened 2019-03-28
FMDB/SQLCipher 3.4.2 to 4.1.0 migration issue


chappanagavendra:
Hi ,

I am using FMDB/SQLCipher for encrypting my database. I have an old version of the application running on the 3.4.2 version of SQLCipher. After the i updated pod , now i am  trying to install a new update(which is running on the 4.x version of SQLCipher), we got the file is not a database error.

Please suggest me

jlcool:
Yeah, what to do? How to change it to 3.x?

jlcool:
I think I've found a way.You can try this modification.

  s.dependency 'FMDB/SQLCipher'
  s.dependency 'SQLCipher', '~>3.4.2'

R4N:
@chappanagavendra @jlcool The default settings have changed in SQLCipher 4, so they will not open a SQLCipher 3 database by default.  There's additional information along with migration options/examples here: https://discuss.zetetic.net/t/upgrading-to-sqlcipher-4/3283

Pratik-Sodha:
I'm facing similar issue. Did anyone found solution ? 

```
  pod 'FMDB/SQLCipher', '~> 2.7'
  pod 'FMDBMigrationManager', '~> 1.4.x'
```
I'm using above pods. And it's working fine till now with following pod version.

FMDB - 2.7.2
SQLCipher - 3.4.1
FMDBMigrationManager - 1.4.1

After pod update i'm getting error database migration fails.

Any help appreciated.

R4N:
@Pratik-Sodha 

The default encryption settings between major versions of SQLCipher are not compatible out of the box (i.e. you can't open a database created with SQLCipher 2.x using SQLCipher 3.x without migrating it or manually modifying the default encryption settings).  You'll need to migrate your 2.x SQLCipher database to use the 3.x encryption settings using [PRAGMA cipher_migrate](https://www.zetetic.net/sqlcipher/sqlcipher-api/#cipher_migrate)

Pratik-Sodha:
@R4N Thank you for feedback. 

But needed version _ dependency_ support into podsepc.

R4N:
@Pratik-Sodha 

I believe you're requesting to add a version to the SQLCipher subspec, is that correct?

The current subspec uses the latest available version of SQLCipher which could include breaking encryption settings changes (if going from one major version to another) as mentioned in this github issue.  

[...]

PR #753, opened 2019-08-21
Adds SQLCipher subspec major version cap which will prevent `pod upda…

R4N:
…te` from updating beyond the current major version until explicitly changed in pod subspec.

Related to: https://github.com/ccgus/fmdb/issues/736

This will maintain a dependency version for SQLCipher subspec so that it won't automatically move to the next major version (if one exists) on folks when they run `pod update`.  This will need to be manually adjusted when we want to bump to the next major version of SQLCipher (i.e. `ss.dependency 'SQLCipher', '~> 5.0'`)


ccgus:
Thanks!

PR #768, opened 2019-12-13
Check that we are actually connected to an SQLCipher DB

niveus:
This will add a check to make sure we're connected to a SQCipher Database. Here is the [advisory](https://discuss.zetetic.net/t/important-advisory-sqlcipher-with-xcode-8-and-new-sdks/1688) regarding the change.

### DIFF
--- a/FMDB.podspec
+++ b/FMDB.podspec
@@ -43,7 +43,7 @@ Pod::Spec.new do |s|
     ss.dependency 'SQLCipher', '~> 4.0'
     ss.source_files = 'src/fmdb/FM*.{h,m}'
     ss.exclude_files = 'src/fmdb.m'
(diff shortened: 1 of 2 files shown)


ccgus:
Do you know if this works for other encryption libraries (such as the one sqlite provides?) Looks like cipher_version might be SQLCipher specific. Are there other ifdefs that can be used to detect if SQLCipher is being used?

niveus:
It seems SQLCipher doesn't implement any project wide CFLAGS ([only specific to itself](https://github.com/sqlcipher/sqlcipher/blob/master/SQLCipher.podspec.json)). I can add one to the FMDB podspec, though, if that will work for you.

ccgus:
That works. Thanks.

Issue #630, opened 2017-12-05
Non-portable path to file '<fmdb/FMDB.h>'; specified path differs in case from file name on disk


Whirlwind:
I have a pod :
```
#import <FMDB/FMDB.h>
```

I got a warning:
```
In file included from 
DDJSONModel+SQLite.m:9:
DDJSONModel+SQLite.h:10:9: warning: non-portable path to file '<fmdb/FMDB.h>'; specified path differs in case from file name on disk [-Wnonportable-include-path]
#import <FMDB/FMDB.h>
        ^~~~~~~~~~~~~
        <fmdb/FMDB.h>
```
I check every settings, I found the pod path `Pods/Headers/Public/FMDB/FMDB.h -> Pods/FMDB/src/fmdb/FMDB.h`

I don't know why the pod search the `Pods/FMDB/src/fmdb/FMDB.h`, not `Pods/Headers/Public/FMDB/FMDB.h`.
I change the folder name `Pods/FMDB/src/fmdb/FMDB.h -> Pods/FMDB/src/FMDB/FMDB.h`, the `fmdb` to `FMDB`, then the warning gone.

ccgus:
Well, your pod needs to be updated to the lowercase version.

Whirlwind:
I can't. There is not a lowercase version in my pod repos.
```
[!] Unable to find a specification for `fmdb` depended upon by `Podfile`
```

I think it may be `Insensitive` in mac osx.

Whirlwind:
Could you rename the folder from `fmdb` to `FMDB` or other?

bamx23:
Looks like the problem can be solved with `spec.header_dir = 'FMDB'` in podspec.
https://guides.cocoapods.org/syntax/podspec.html#header_dir

PR #774, opened 2020-02-18
Fix header_dir path in podspec

bamx23:
As far as I can see from other headers it's expected to use `#import <fmdb/FMDB.h>`, and **not** `#import <FMDB/FMDB.h>`.
https://github.com/ccgus/fmdb/blob/9a98636848905f48b6b9a304f2b90f9b3c2c9c4a/src/extra/fts3/FMDatabase%2BFTS3.h#L9

In some cases current podspec may not work as expected. This change will ensure the consistency in directory name register.

Related to #630


ccgus:
Thanks!

PR #787, opened 2020-05-05
Improve FMDBVersion

robertmryan:
- Previously, `FMDBVersion` would break if any component exceeded 9.
- With this change, it doesn't break until any component exceeds 15, at which point it will max out at 0xf for that component and log an error.
- Added documentation suggesting FMDBUserVersion with NSNumericSearch option.
- Added unit tests for proper version number checking.
- While unit testing, I noticed that `testBoolInsert` was attempting `[NSNumber numberWithBool:12]`, which is not valid. Replaced with `YES`.


robertmryan:
FWIW, I debated changing `FMDBVersion` to return `-1` if the returned numeric value was invalid because a component version number exceeded 15. That seems like the correct behavior. And it seems like we should just deprecate this method because of this fundamental problem with this hexadecimal version numbering representation. But I’ll defer to you, @ccgus.

But at least this update defers the problem. The previous rendition was going to break when you hit, for example, 2.10.0. This rendition doesn’t break until 2.16.0.

src/fmdb/FMDatabase.h
```diff
@@ -1085,11 +1085,36 @@ typedef NS_ENUM(int, FMDBCheckpointMode) {
...
+
+ @warning This routine will not work if any component of the version number exceeds 15.
+       For example, if it is version 2.17.3, this will max out at 0x2f3.
+       For this reason, we would recommend using `FMDBUserVersion` and with `NSNumericSearch` option, e.g.
+
+ @code
+ NSComparisonResult result = [[FMDatabase FMDBUserVersion] compare:@"2.11.0" options:NSNumericSearch];
+ @endcode
+
+ @returns The version number in hexadecimal, e.g., 0x0243 for version 2.4.3. If any component exceeds what can be
```
robertmryan, line 1114: Whoops, I didn’t finish that comment. I’ll update and post additional commit (which you should feel free to squash).

robertmryan: Fixed.

robertmryan:
I didn’t mean to update master. I’ll make sure to post future PR as a separate branch. Sorry about that.

PR #788, opened 2020-05-06
Allow rebinding

robertmryan:
* Allow re-binding of SQLite statements

  - Reduced `executeUpdate` to call `executeQuery` and remove redundant code;
  - Abstracted binding code out of from `executeQuery`;
  - Created “prepare” method on `FMDatabase` (that prepares statement but doesn’t bind anything);
  - Created “bind” methods on `FMResultSet` to bind values; and
  - Created unit tests for both array and dictionary parameter types; each tests both update queries and select queries.

Usage:

For execute queries:

```
FMResultSet *rs = [db prepare:@"INSERT INTO foo (bar) VALUES (?)"];
if ([rs bindWithArray:@[value1]]) {
     if ([rs step]) {
         // successful
     }
}

if ([rs bindWithArray:@[value2]]) {
    if ([rs step]) {
        // successful
    }
}
```

While it’s most useful for `INSERT` and `UPDATE` queries, you can theoretically use it for select queries, too:

```
rs = [db prepare:@"SELECT bar FROM foo WHERE bar = ?"];
if ([rs bindWithArray:@[value1]]) {
    while ([rs next]) {
        NSString *value = [rs stringForColumnIndex:0];
    }
}

if ([rs bindWithArray:@[value2]]) {
    while ([rs next]) {
        NSString *value = [rs stringForColumnIndex:0];
    }
}
```

* Detect and handle `sqlite3_bind` errors;

  - Create unit test that manifests bind errors (by trying to insert `NSData` that is too big); it checks this limit and tries to insert something too big;
  - Created `limitFor:value:` to call `sqlite3_limit` so I could create aforementioned test that would reliably fail; and
  - Modified bind routines to return result of `sqlite3_bind_xxx` failures.

* While looking at binding code, I noticed that it was using `SQLITE_STATIC`. That’s a little dangerous depending upon the calling point:

  - Create unit test that manifests when using `SQLITE_STATIC` (it’s a contrived example, but illustrates the problem); and
  - Changed binding routines to use `SQLITE_TRANSIENT` not `SQLITE_STATIC`.

* In a separate commit, I replaced all the http references to sqlite.org with https reference, as the links no longer work without https.

### DIFF
--- a/CHANGES_AND_TODO_LIST.txt
+++ b/CHANGES_AND_TODO_LIST.txt
@@ -3,8 +3,11 @@ Zip, nada, zilch.  Got any ideas?
 
 If you would like to contribute some code ... awesome!  I just ask that you make it conform to the coding conventions already set in here, and to add the necessary of tests for your new code to tests target.  And of course, the code should be of general use to more than just a couple of folks.  Send your patches to gus@flyingmeat.com.
 
+2020.05.06 Version 2.7.7
+    Add `prepare` and `bind` methods so you can prepare a statement once and bind values repeatedly.
+
 2020.04.23 Version 2.7.6
-A new tag for the Swift Package Manager.
+    A new tag for the Swift Package Manager.
 
 2018.10.23 Version 2.7.5
     Xcode 10 support. Probably some other stuff over the past year as well.
--- a/README.markdown
+++ b/README.markdown
@@ -3,15 +3,15 @@
 [![CocoaPods Compatible](https://img.shields.io/cocoapods/v/FMDB.svg)](https://img.shields.io/cocoapods/v/FMDB.svg)
 [![Carthage Compatible](https://img.shields.io/badge/Carthage-compatible-4BC51D.svg?style=flat)](https://github.com/Carthage/Carthage)
 
-This is an Objective-C wrapper around SQLite: http://sqlite.org/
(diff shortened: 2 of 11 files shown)


robertmryan:
A couple of design questions:

* I decided to use `step` to perform `sqlite3_step`, returning `true` if `SQLITE_DONE` and `false` if not. That stands in contrast to `next` which returns `true` if `SQLITE_ROW` and `false` if not. Is that too confusing?

* You theoretically can add variadic rendition of `bind`. I only added array and dictionary renditions of `bind`.

robertmryan:
Also, when fixing the `README`, my goal was to fix the various broken links. There is was one lingering broken reference to `www.versioneye.com` in the Cocoapods section. It was not immediately clear what the correct link was there, so I just removed it for now. Obviously, feel free to add back a working link.

ccgus:
Oh hey, good idea having `executeUpdate:` just call into `executeQuery:`! It all looks pretty good, thanks for doing this!

ccgus:
For the variadic version of bind- maybe we'll do that if anyone requests it? I did notice it was missing, but I wasn't about to switch my code over to use it…

I don't think the difference between step and next is a big deal. They are used in different manners, so it's reasonable to have different behaviors.

PR #789, opened 2020-05-08
Uninitialized object in manual reference counting

robertmryan:
I can’t imagine that anyone is using manual referencing counting any more, but there was an uninitialized pointer. 

Screen Shot 2020-05-07 at 6 59 28 PM

Turns out that `blockError` reference is not needed, anyway, so let’s cut the Gordian knot.


ccgus:
👍

PR #794, opened 2020-05-27
Added SPM section to README

MartinJNash:
Added section about importing FMDB via the Swift Package Manager to README.md


ccgus:
Thanks Martin! (Also, long time no see!)

PR #806, opened 2020-11-30
Update FMDB.podspec

jamesstout:
added osx.deployment_target  = '10.11'

### DIFF
--- a/FMDB.podspec
+++ b/FMDB.podspec
@@ -7,6 +7,7 @@ Pod::Spec.new do |s|
   s.author = { 'August Mueller' => 'gus@flyingmeat.com' }
   s.source = { :git => 'https://github.com/ccgus/fmdb.git', :tag => "#{s.version}" }
   s.requires_arc = true
+  s.osx.deployment_target  = '10.11'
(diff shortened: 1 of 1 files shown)


Kaspik:
@ccgus Based on https://github.com/ccgus/fmdb/pull/743 we think minimum deployment target of this project is 10.11. But this `required` value is not mentioned in `podspec`. This results to all clients implementing `post_install` hook (see https://github.com/CocoaPods/CocoaPods/pull/10070 and https://github.com/CocoaPods/CocoaPods/issues/9884).

Please take a look and merge in, thank you! :)

ccgus:
Thanks! (I don't use CocoaPods, but this looks reasonable so I'm merging)

Kaspik:
@ccgus Thanks!

On similar note, your `SPM` `Package.swift` is missing completely the same :)

```
platforms: [
        .macOS(.v10_11),
    ],
```
https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html

jamesstout:
Many thanks.

PR #808, opened 2020-12-22
Fix Swift Package Manager support

woxtu:
Hi,

The current release installed via SPM cannot import headers as follows:

```objective-c
#import <FMDB/FMDB.h>
```

This PR creates the public headers directory to fix that.


ccgus:
Why not import them via #import "FMDB.h", or just what is needed?

Jeehut:
I import via `@import FMDB;`, did you try that.

GLinnik21:
> Why not import them via #import "FMDB.h", or just what is needed?

I think because the `#import' syntax is inconsistent between SPM and CocoaPods.

PR #810, opened 2021-01-02
Fixed issue from [3e5efdf] (pull request #806) which broke cocoapods for iOS only projects

evands:
If specifying an OS X deployment ([3e5efdf]), an iOS one needs to be specified, too.


incanus:
Obsolete CocoaPods-related after https://github.com/ccgus/fmdb/issues/860.

PR #815, opened 2021-03-11
Add null check before sqlite3_step

zsteed:
Hello, I want to add a null check on `_statement` just before we add it in the sqlite3_step.

Our production app is getting a lot of crash's at this line `int rc = sqlite3_step([_statement statement]);`

It is random and I have no idea how to re-create this. 

I'm guessing it is some sort of race condition where the _statement is set to null in the `close()` method.

I just want to avoid a crash.

Let me know what you guys think.


ccgus:
The statement can have a bad value without being null. I'd suggest posting a stack trace of your crashes, with all threads showing.

Issue #73, opened 2012-05-29
Zero-length NSData parameter binds to SQL null value


snej:
A zero-length NSData object passed as a parameter binding for a SQL statement results in a SQL 'null' value, not a zero-length blob as expected.

``` objective-c
    // Test case: This assertion fails.
    NSAssert(([_fmdb boolForQuery: @"SELECT ? not null", [NSData data]]), @"Empty data should not be null");
```

This is significant because SQL 'null' values have odd behavior with respect to things like equality comparison. Additionally, my app schema has a blob column where a null value means "there is no data" as distinct from "the data is empty", but this distinction gets lost when the data is saved.

I'm not sure why this happens. The binding is done by the following lines in `-[FMDatabase bindObject:toColumn: inStatement:`:

```
else if ([obj isKindOfClass:[NSData class]]) {
    sqlite3_bind_blob(pStmt, idx, [obj bytes], (int)[obj length], SQLITE_STATIC);
}
```

It may be that `sqlite3_bind_blob` binds a null value if the data pointer is NULL or the length is zero, but that seems like strange behavior since there is already a `sqlite3_bind_null` function.

ArturFriesen:
Is it possible to integrate this fix? I have the same problem like the issue creator.

ccgus:
@ArturFriesen If there's some tests and a pull request, I can look at it.

robertmryan:
> It may be that `sqlite3_bind_blob` binds a null value if the ... the length is zero, ...

That is precisely what SQLite does. If the length is zero (regardless of the address), when you call `sqlite3_bind_blob`, then the subsequent call to `sqlite3_column_blob` will return `NULL`. 

This is not a FMDB problem, but rather a “feature” of SQLite. 

You go on to ask:

> ... but that seems like strange behavior since there is already a `sqlite3_bind_null` function.
>
> ... Additionally, my app schema has a blob column where a null value means "there is no data" as distinct from "the data is empty", but this distinction gets lost when the data is saved.

There actually is a minor difference between calling `sqlite3_bind_null` and `sqlite3_bind_blob` with a zero length. Sure, in both cases, `sqlite3_column_blob` will return `NULL`, but `sqlite3_column_type` will return `SQLITE_NULL` for the former and `SQLITE_BLOB` for the latter. So, SQLite knows whether it was a zero length BLOB or a `NULL`.

- - -

[...]

PR #818, opened 2021-03-25
Add typeForColumn functions

robertmryan:
* Add `typeForColumn` and `typeForColumnIndex` methods and tests
* Update warnings in `dataForColumn` and `dataForColumnIndex` to describe SQLite behavior re zero-length blobs
* Bump version to 2.7.8

This should permit accurate handling of zero-length BLOBs, as discussed in #73.

So, for example, if you want to delineate between NULL and a zero length BLOB, in Swift:

```swift
let fileURL = URL(fileURLWithPath: NSTemporaryDirectory())
    .appendingPathComponent("test.sqlite")

do {
    let db = FMDatabase(url: fileURL)
    db.open()
    try db.executeUpdate("CREATE TABLE test (data BLOB)", values: nil)
    let data = "foo".data(using: .utf8)!
    let sql = "INSERT INTO test (data) VALUES (?)"
    try db.executeUpdate(sql, values: [data])
    try db.executeUpdate(sql, values: [Data()])
    try db.executeUpdate(sql, values: [NSNull()])

    let rs = try db.executeQuery("SELECT * FROM test", values: nil)
    while rs.next() {
        let type = rs.type(forColumnIndex: 0)
        switch type {
        case .blob:
            if let data = rs.data(forColumnIndex: 0) {
                print("blob with \(data.count) bytes")
            } else {
                print("blob but no data")
            }

        case .null:
            print("Null")

        default:
            print("type is \(type)")
        }
    }
} catch {
    print(error)
}
```

That will produce:

```
blob with 3 bytes
blob but no data
Null
```

### DIFF
--- a/CHANGES_AND_TODO_LIST.txt
+++ b/CHANGES_AND_TODO_LIST.txt
@@ -3,6 +3,9 @@ Zip, nada, zilch.  Got any ideas?
 
 If you would like to contribute some code ... awesome!  I just ask that you make it conform to the coding conventions already set in here, and to add the necessary of tests for your new code to tests target.  And of course, the code should be of general use to more than just a couple of folks.  Send your patches to gus@flyingmeat.com.
 
+2020.03.25 Version 2.7.8
+    Add `valueForColumn` functions to expose `sqlite3_column_type`. Add comments to `dataForColumn` re SQL behavior of zero-length BLOBs.
+
 2020.05.06 Version 2.7.7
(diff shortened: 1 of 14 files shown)


robertmryan:
FWIW, I have also removed the `testVersionNumber` test, as that is testing a deprecated method. In Xcode 12.4, at least, calling deprecated methods is now an error, rather than just a warning.

ccgus:
Danke!

Issue #793, opened 2020-05-13
Update podspec?


robertmryan:
I published version 2.7.7 release (making it available for Carthage users), but when I went to push the Cocoapods trunk, I got a whole bunch of warnings:

```none
Updating spec repo `trunk`
Validating podspec
nil versions are discouraged and will be deprecated in Rubygems 4
 -> FMDB (2.7.7)
    - NOTE  | [FMDB/standard, FMDB/FTS, FMDB/standalone, and more...] xcodebuild:  note: Using new build system
    - NOTE  | [FMDB/standard, FMDB/FTS, FMDB/standalone, and more...] xcodebuild:  note: Building targets in parallel
    - NOTE  | [FMDB/standard, FMDB/FTS, FMDB/standalone, and more...] xcodebuild:  note: Planning build
    - NOTE  | [FMDB/standard, FMDB/FTS, FMDB/standalone, and more...] xcodebuild:  note: Constructing build description
    - WARN  | [FMDB/standard, FMDB/FTS, FMDB/standalone, and more...] xcodebuild:  FMDB/src/fmdb/FMDatabaseQueue.m:101:9: warning: 'dispatch_queue_set_specific' is only available on macOS 10.7 or newer [-Wunguarded-availability]
    - NOTE  | [FMDB/standard, FMDB/FTS, FMDB/standalone, and more...] xcodebuild:  /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/usr/include/dispatch/queue.h:1498:1: note: 'dispatch_queue_set_specific' has been marked as being introduced in macOS 10.7 here, but the deployment target is macOS 10.6.0
    - NOTE  | [FMDB/standard, FMDB/FTS, FMDB/standalone, and more...] xcodebuild:  FMDB/src/fmdb/FMDatabaseQueue.m:101:9: note: enclose 'dispatch_queue_set_specific' in an @available check to silence this warning
    - WARN  | [FMDB/standard, FMDB/FTS, FMDB/standalone, and more...] xcodebuild:  FMDB/src/fmdb/FMDatabaseQueue.m:184:54: warning: 'dispatch_get_specific' is only available on macOS 10.7 or newer [-Wunguarded-availability]
[...]

PR #824, opened 2021-06-12
Fix deployment targets and CocoaPods integration

tstump-phunware:
**Background**
As of today, the latest FMDB release available through the CocoaPods CDN specs repo is version `2.7.5`. The reason why version `2.7.7` is not available is because the FMDB podspec fails the CocoaPods [required] lint step (ref: #793). This can impact the visibility and contribution to future FMDB releases in the context of developers who rely solely on public releases provided by the CocoaPods CDN specs repo.

**Purpose**
The purpose of these changes is to sync up the iOS, Mac OS X, and Watch OS platform deployment targets within both the Xcode project file and CocoaPods podspec, so that subsequent FMDB releases (starting with version `2.7.8`) can be made available through the CocoaPods CDN specs repo.

**Attachments**
- [fmdb-2.7.7-podspec-validation.txt](https://github.com/ccgus/fmdb/files/6641761/fmdb-2.7.7-podspec-validation.txt)

**Related**
- Issue #793
- PR #810
- PR #806


fmdb.xcodeproj/project.pbxproj
```diff
@@ -620,14 +620,10 @@
 				};
 			};
 			buildConfigurationList = 1DEB927808733DD40010E9CD /* Build configuration list for PBXProject "fmdb" */;
-			compatibilityVersion = "Xcode 8.0";
-			developmentRegion = English;
+			compatibilityVersion = "Xcode 12.0";
```
tstump-phunware, line 623: Bumped the compatibility since Apple requires all submissions to be built using Xcode 12 or newer as of April 21 (ref: https://developer.apple.com/news/?id=ib31uj1j)

fmdb.xcodeproj/project.pbxproj
```diff
@@ -917,7 +918,10 @@
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_PARAMETER = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
+				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
```
tstump-phunware, line 921: Lines 921, 922, and 923:

Moved the deployment targets to be defined at the project level so its easier to spot compilation issues when considering a deployment target bump. IMHO: This also makes it easier to maintain, as the project's targets simply inherit their deployment target from a single version of truth.

tstump-phunware:
Ready for review.

cc: @ccgus, @robertmryan

ccgus:
Thanks for doing this!

Issue #522, opened 2016-08-29
Unsafe unretained can cause a crash


ramakula:
Hello @ccgus,
Is there any specific reason why you have used __unsafe_unretained instead of weak in FMDatabasePool.h? __unsafe_unretained can lead to dangling pointer. Here is what I think: "If you try to access an unowned reference after the instance that it references is deallocated, you will trigger a runtime error. Use unowned references only when you are sure that the reference will always refer to an instance.".  
<img width="421" alt="screen shot 2016-08-29 at 2 59 08 pm" src="https://cloud.githubusercontent.com/assets/2745769/18069124/d8ac1b92-6df9-11e6-9353-a331e6e9c265.png">

ccgus:
Probably because I wrote these bits before weak was introduced.

ramakula:
Ah makes sense. I can go ahead and submit a pull request. Let me know.

ccgus:
I've got it tagged for 3.0 so if you want to make the change, you can do it there. I'm not going to make a behavior change like that in the 2.x cycle (even though it's _probably_ safe to do. It's the _probably_ part that worries me)

ramakula:
Got it

PR #836, opened 2022-10-06
Fix dangling reference warning on assign object property

tiwoc:
[This doesn't change the behavior](https://clang.llvm.org/docs/AutomaticReferenceCounting.html#property-declarations), it just makes the behavior explicit and silences this compiler warning when using FMDB in a project that has the `-Wobjc-property-assign-on-object-type` warning flag:

> 'assign' property of object type may become a dangling reference; consider using 'unsafe_unretained'

The alternative would have been to make this a weak property, but this would have been a behavior change, and after reading the discussion https://github.com/ccgus/fmdb/issues/522, I gather that this is something that shouldn't be done in FMDB without bumping the major version number.

### DIFF
--- a/src/fmdb/FMDatabasePool.h
+++ b/src/fmdb/FMDatabasePool.h
@@ -38,7 +38,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 /** Delegate object */
 
-@property (atomic, assign, nullable) id delegate;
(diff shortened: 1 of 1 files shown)


lzcuriosity:
您的邮件已经收到，我会尽快查阅，谢谢！Best Wishes!

Issue #844, opened 2023-04-18
Error with Xcode 14.3 because of deployment_target (9.0)


gerken-tss:
I am using `fmdb` as a transitive dependency of a Flutter package ([sqflite](https://pub.dev/packages/sqflite)).

Whenever I try to compile the application, Xcode complains about FMDB pod with this error:

>Error (Xcode): File not found: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/arc/libarclite_iphonesimulator.a
>
>Error (Xcode): Linker command failed with exit code 1 (use -v to see invocation)
>
>Could not build the application for the simulator.
>Error launching application on iPhone 14 Pro Max.

According to [SO](https://stackoverflow.com/questions/75574268/missing-file-libarclite-iphoneos-a-xcode-14-3) this issue is connected to `deployment_target` being too low (requires 11.0 at least).

I'm hesitating to accept the solution being mentioned in the SO post as it uncontrollably overrides all the `deployment_target` properties of every pod (this doesn't seem right).

A better solution would be for the author every pod with outdated `deployment_target` to willingly increase it.

kyle-seongwoo-jun:
same problem here

Yck-Dakucha:
same

PR #849, opened 2023-09-08
Ups podspec min versions for modern xcode (xcode 14.3.x)

R4N:
iOS min 11, macOS min 10.13, watchOS min 7.0.

Should fix: https://github.com/ccgus/fmdb/issues/844

### DIFF
--- a/FMDB.podspec
+++ b/FMDB.podspec
@@ -7,9 +7,9 @@ Pod::Spec.new do |s|
   s.author = { 'August Mueller' => 'gus@flyingmeat.com' }
   s.source = { :git => 'https://github.com/ccgus/fmdb.git', :tag => "#{s.version}" }
   s.requires_arc = true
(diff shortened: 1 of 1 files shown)


ccgus:
I think you meant '13' instead of '10.13' for the MacOS min? Or does podspec need it to be in a 10.x format? (Also - maybe make it '11' since I know folks still target older releases. MacOS 13 isn't even a year old at this point.

R4N:
Deployment targets are the _min_ os version capable of running it. I believe we'll want to leave this as low as possible while still supporting modern Xcode version to allow as many devices to be able to use it as possible. For macOS it needs to use the full macOS version number (i.e. 10.13 -- High Sierra). When creating a new project in Xcode 14.3.x the minimum deployments available are what the Podspec is adjusted to in this pull request.

PR #851, opened 2023-10-31
Added Support for VisionOS

medabida:
create a new target FMDB xrOS to add support for Building FMDB for VisionOS


ccgus:
Thanks for doing this!

